PR #1, opened 2019-04-29
Dockerize dev environment

GiancarloFusiello:
Adds configuration files to run airflow, postgres and s3 locally.


docker-compose.yml
```diff
@@ -0,0 +1,39 @@
+version: "3.7"
+
+services:
+  localstack-s3:
+    image: localstack/localstack:latest
```
giorgiosironi: this is a huge image, but I haven't found a better option with a liberal license

giorgiosironi: do they have any good released version more stable than `latest`?

docker-compose.yml
```diff
@@ -0,0 +1,39 @@
+version: "3.7"
```
giorgiosironi: Travis CI has a default `docker-compose` that allows up to `3.4`, compare https://github.com/libero/sample-configuration/blob/master/docker-compose.yml#
A newer version requires to update `docker-compose` on every build, but it doesn't look like there is syntax in here that requires a very new version

Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM python:3.7.3-slim
+
+ENV PYTHONUNBUFFERED=1
+ENV PYTHONDONTWRITEBYTECODE=1
+ENV AIRFLOW_HOME=/src/airflow
+
+WORKDIR /src/app
```
giorgiosironi: `/srv/app`?

giorgiosironi: (there is a `/usr/src` in https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard but it's mostly kernel C source code)

docker-compose.yml
GiancarloFusiello: updated

Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+COPY ./requirements ./requirements
+
+RUN pip install -U pip \
+    && set -ex \
+    && apt-get update -yq \
+    && apt-get install -yq --no-install-recommends build-essential \
+    && pip install --no-cache-dir -r ./requirements/base.txt \
+    && apt-get remove --purge --autoremove -yq build-essential \
+    && rm -rf ~/.cache/* \
+    && rm -rf /tmp/*
```
giorgiosironi: include `/var/lib/apt/lists/*` to remove the package indexes created by `apt-get update`

docker-compose.yml
GiancarloFusiello: Doesn't appear to be. `latest` is the latest version of the image.

Dockerfile
GiancarloFusiello: updated

Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+ARG env
```
giorgiosironi: immutable images should be reusable across environments, so this should be a fixed set of choices (usually two, `dev` and non-`dev`) more than a string with open options; in `search` this is called [`ARG dev`](https://github.com/libero/search/blob/master/docker/Dockerfile#L30), but it's used in a command line option

Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+    && apt-get remove --purge --autoremove -yq build-essential \
+    && rm -rf ~/.cache/* \
+    && rm -rf /var/lib/apt/lists/* \
+    && rm -rf /tmp/*
+
+ARG env
+RUN pip install --no-cache-dir -r ./requirements/${env}.txt \
+    && rm -rf /tmp/*
+
+EXPOSE 8080
```
giorgiosironi: reason for it not being `80`? The host can always forward 8080 to the container's 80, but maybe 80 is being reserved for something else.

Makefile
```diff
@@ -0,0 +1,13 @@
+help:
+	@echo "start - starts all services"
+	@echo "stop  - stops all services and removes anonymous volumes"
+	@echo "shell - runs an instance of the 'app' service and presents a bash prompt"
+
+start:
+	docker-compose up
```
giorgiosironi: consider an explicit `--build` here

docker-compose.yml
```diff
@@ -0,0 +1,39 @@
+version: "3.4"
+
+services:
+  localstack-s3:
+    image: localstack/localstack:latest
+    container_name: localstack-s3
```
giorgiosironi: this shouldn't be needed in general, as the service name `localstack-s3` is used to refer to the container's hostname from any other container

docker-compose.yml
```diff
@@ -0,0 +1,39 @@
...
+    environment:
+      - SERVICES=s3:5002
```
giorgiosironi: worth to use the [standard `localstack` port for S3](https://github.com/localstack/localstack#overview) if free

docker-compose.yml
```diff
@@ -0,0 +1,39 @@
...
+      - PORT_WEB_UI=8081
+      - DEFAULT_REGION=eu-west-2
+      - DATA_DIR=/tmp/localstack/data
+    ports:
+      - "5002:5002"
+      - "8081:8081"
+
+  db:
+    image: postgres
```
giorgiosironi: `postgres:10.7` or `postgres:9.6` or similar can be more precise

Dockerfile
GiancarloFusiello: updated so that `/src/airflow` is the working directory.

docker-compose.yml
```diff
@@ -0,0 +1,39 @@
...
+  app:
+    build:
+      dockerfile: Dockerfile
+      context: .
+      args:
+        - env=dev
+    volumes:
+      - ./airflow.cfg:/src/airflow/airflow.cfg
+    ports:
+      - 8080:8080
```
giorgiosironi: any `healthcheck` here?

Dockerfile
GiancarloFusiello: I was just making airflow available to the host browser. I have also done the same for localstack on port 8081. I think it keeps the urls consistent by having one at 8080 and the other at 8081.

Makefile
```diff
@@ -0,0 +1,13 @@
...
+
+stop:
```
giorgiosironi: A bit asymmetric wrt `docker-compose stop`, which stops the containers but doesn't destroy the state (the volumes)

Makefile
GiancarloFusiello: I had but couldn't find a reason for it at this moment.

docker-compose.yml
```diff
@@ -0,0 +1,39 @@
...
+      retries: 3
+
+  app:
+    build:
+      dockerfile: Dockerfile
+      context: .
+      args:
+        - env=dev
+    volumes:
+      - ./airflow.cfg:/src/airflow/airflow.cfg
```
giorgiosironi: a sizable file to provide from outside; any options for making it smaller or shipping a default `.cfg` file in the image? Not sure how many of the values in there are local (paths to folders) and stable vs. remote (host names that will change across environments)

docker-compose.yml
GiancarloFusiello: removed.

docker-compose.yml
GiancarloFusiello: updated.

Makefile
GiancarloFusiello: The make target is stop. The command is `docker-compose down -v`.

docker-compose.yml
GiancarloFusiello: This is automatically generated. I'm only changing a few values. An alternative is to use environment variables to set configuration values. I could try removing all of the values I'm not using but I don't what the effects of this would be.

docker-compose.yml
GiancarloFusiello: do you have a suggestion for checking the health status of this container?

giorgiosironi: `curl localhost:8080/`? Assuming `curl` or `wget` or similar is available, and `/` is an existing path - to be verified.

Makefile
giorgiosironi: Yes, `stop` doesn't just stop the containers with this command but also deletes data; it's local so it should not be important state...

docker-compose.yml
giorgiosironi: Main value that stands out is the database connection parameters:
https://github.com/libero/elife-style-content-adapter-prototype/pull/1/files#diff-ea36df1840fe05258ac3690974a20733R53
which means the file cannot be shipped inside the image. The existing standard is to put these configuration files into a [`.docker/` subfolder](https://github.com/libero/browser/tree/master/.docker) so that it's clear what is needed to [run the image elsewhere in a non-`dev` environment](https://github.com/libero/sample-configuration).

Makefile
giorgiosironi: The main difference is `make start` won't see changes to the files in the image (requiring a separate `docker-compose build && make start`) but the only ones that this applies to at the moment are `requirements/*` files.

Dockerfile
giorgiosironi: :+1: for consistency between host and container port in local development

Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM python:3.7.3-slim
+
+ENV PYTHONUNBUFFERED=1
+ENV PYTHONDONTWRITEBYTECODE=1
+ENV AIRFLOW_HOME=/src/airflow
```
giorgiosironi: `src` indicates source files in Unix speak, but there are installed libraries, logs, etc. in here; go for just `/airflow` like other applications (or [`/app`](https://github.com/libero/browser/blob/master/Dockerfile#L31))

docker-compose.yml
GiancarloFusiello: updated.

docker-compose.yml
GiancarloFusiello: Bear in mind the title of the PR: `Dockerize dev environment`. I'll update this to use environment variables which will only be present in the dev image.

Dockerfile
GiancarloFusiello: updated to /airflow.

docker-compose.yml
GiancarloFusiello: updated.

Makefile
GiancarloFusiello: Would you like me to remove `-v`?

docker-compose.yml
GiancarloFusiello: `curl` or `wget` are not installed by default. Would you like me to add them to add a health check? should I add a health check to `localstack` also?

Dockerfile
GiancarloFusiello: I've updated the docker file to use multistage build so that everything is explicit.

docker-compose.yml
giorgiosironi: > I'll update this to use environment variables 

That's good, now that I am aware of the possibility.

Makefile
giorgiosironi: Between the two options of 1) named `stop` but it deletes data and 2) leftover data from previous runs, I'd go for 1) for a development environment (leaving as it is).

docker-compose.yml
giorgiosironi: The `app` seems designed to wait for the other services, so it's the only one where I see the healthcheck as strictly necessary so that we can deploy, [wait for the container to be healthy](https://github.com/libero/scripts/blob/master/docker/wait-healthy.sh) and run smoke tests. Depending on the application, sometimes checking the port is open with `netcat` is enough; sometimes a real HTTP request is needed which would mean installing the missing `curl`/`wget`.

Dockerfile
```diff
@@ -0,0 +1,41 @@
+FROM python:3.7.3-slim as base
+
+# change /src/app to /airflow
+ENV AIRFLOW_HOME=/airflow
+ENV AIRFLOW__CORE__DAGS_FOLDER=${AIRFLOW_HOME}/dags
+ENV AIRFLOW__CORE__BASE_LOG_FOLDER=${AIRFLOW_HOME}/logs
```
giorgiosironi: so the configuration file is gone, good news

Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+RUN pip install --no-cache-dir -r ./requirements/dev.txt \
+    && rm -rf ~/.cache/* \
+    && rm -rf /tmp/*
+
+ADD https://github.com/ufoscout/docker-compose-wait/releases/download/2.5.0/wait /wait
+RUN chmod +x /wait
+
+EXPOSE 8080
+
+CMD /wait && airflow initdb && airflow webserver -p 8080 && airflow scheduler
```
giorgiosironi: would check if `airflow webserver` is asynchronous and returns? Otherwise `airflow scheduler` wouldn't get executed. If the intention is to have to separate processes they are usually implemented with two containers running the same image

GiancarloFusiello: I'm unsure of the implementation but this is what the `airflow` documentation says to do which I think is fine for development purposes, especially for a prototype.

giorgiosironi: This command doesn't seem to be running `airflow scheduler`:
```
 3863 ?        Sl     0:00  \_ containerd-shim -namespace moby -workdir /var/lib/containerd/io.containerd.runtime.v1.linux/moby/808de0303907108b43b141a4fc8ce1aa55a85437f05c18db2462d80ab752da42 -address /run/containerd/containerd.sock -containerd-binary /usr/bin/containerd -runtime-root /var/run/docker/runtime-runc
 3897 ?        Ss     0:00      \_ /bin/sh -c /wait && airflow initdb && airflow webserver -p 8080 && airflow scheduler
 4199 ?        S      0:01          \_ /usr/local/bin/python /usr/local/bin/airflow webserver -p 8080
 4275 ?        S      0:01              \_ gunicorn: master [airflow-webserver]
 4287 ?        Sl     0:00                  \_ [ready] gunicorn: worker [airflow-webserver]
 4290 ?        Sl     0:00                  \_ [ready] gunicorn: worker [airflow-webserver]
 4292 ?        Sl     0:00                  \_ [ready] gunicorn: worker [airflow-webserver]
 4540 ?        Sl     0:00                  \_ [ready] gunicorn: worker [airflow-webserver]
```
because `airflow-webserver -p 8080` (correctly) blocks while it is running.

docker-compose.yml
giorgiosironi: There isn't a `netcat` in the container, so this is the best I have found:
```
bash -c 'echo > /dev/tcp/localhost/8080'
```
It's a bash-ism to check a port is open; assumes if the port has been opened the webserver is ready to answer requests.

Dockerfile
GiancarloFusiello: correct, I've just seen there is a `/health` endpoint indicating this. Giving the official documentation another read.

giorgiosironi:
Current processes tree:
```
5367 ?        Sl     0:00  \_ containerd-shim -namespace moby -workdir /var/lib/containerd/io.containerd.runtime.v1.linux/moby/ab6b31d2dacdbe36d378ee55ab3d618b9a31c39473$b9874a4472879183d894a -address /run/containerd/containerd.sock -containerd-binary /usr/bin/containerd -runtime-root /var/run/docker/runtime-runc
 5402 ?        Ss     0:00      \_ /bin/sh -c /wait && airflow initdb && /bin/sh -c "airflow scheduler -D &" && airflow webserver -p 8080
 5637 ?        S      0:04          \_ /usr/local/bin/python /usr/local/bin/airflow scheduler -D
 5894 ?        Sl     0:00          |   \_ /usr/local/bin/python /usr/local/bin/airflow scheduler -D
 5638 ?        S      0:01          \_ /usr/local/bin/python /usr/local/bin/airflow webserver -p 8080
 5719 ?        S      0:01              \_ gunicorn: master [airflow-webserver]
 5793 ?        Sl     0:00                  \_ [ready] gunicorn: worker [airflow-webserver]
 5794 ?        Sl     0:00                  \_ [ready] gunicorn: worker [airflow-webserver]
 5795 ?        Sl     0:00                  \_ [ready] gunicorn: worker [airflow-webserver]
 5796 ?        Sl     0:00                  \_ [ready] gunicorn: worker [airflow-webserver]
```
All reasonable and multiple processes running as expected. In the medium term, multiple containers running from the same image will be more flexible for isolation/scaling.

airflow_entrypoint.sh
```diff
@@ -0,0 +1,16 @@
...
+    exit
+fi
+
+/wait
+
+if [ "$1" != "initdb" ]; then
+    sleep 5
+fi
+
+airflow $@
```
giorgiosironi: nothing strange in here

airflow_entrypoint.sh
```diff
@@ -0,0 +1,16 @@
...
+
+if [ "$#" -lt 1 ]; then
+    echo "Please add the command you would like to run with Airflow."
+    exit
+fi
+
+/wait
+
+if [ "$1" != "initdb" ]; then
+    sleep 5
```
giorgiosironi: arbitrary sleep is always a smell, what is this waiting for? `/wait` or sequencing the parts of `command:` with `&&` should address all of that?

giorgiosironi: Add a comment explaining why, so that if/when we go to production we can implement a proper waiting condition on `postgres` being initialized with tables.

docker-compose.yml
```diff
@@ -0,0 +1,49 @@
...
+    &airflow
+    build:
+      target: dev
+      context: .
+    command: initdb
+    depends_on:
+      - db
+    environment:
+      WAIT_HOSTS: db:5432
+    healthcheck:
```
giorgiosironi: I'd assume `initdb` exits once finished, so it doesn't need an healthcheck (the other `airflow_*` services do).

airflow_entrypoint.sh
GiancarloFusiello: comments added.

docker-compose.yml
GiancarloFusiello: Correct, it does exit. I added this so the setting is copied to the other services.

giorgiosironi: Unused but it shouldn't necessarily hurt

airflow_healthcheck.py
```diff
@@ -0,0 +1,18 @@
+import json
+import sys
+from urllib.request import urlopen
+
+HEALTHY = 'healthy'
+url = 'http://airflow_webserver:8080/health'
```
giorgiosironi: hardcoded relative URL in image - won't work outside of this `docker-compose` configuration; I suggest to use `localhost` and only have the healthcheck on `webserver`, not putting one on `scheduler` for now

docker-compose.yml
giorgiosironi: See suggestion on https://github.com/libero/elife-style-content-adapter-prototype/pull/1/files#r280103287 which would make this only available on `webserver`

airflow_healthcheck.py
GiancarloFusiello: updated.

docker-compose.yml
GiancarloFusiello: updated.

PR #2, opened 2019-05-02
Add worker to docker dev env

GiancarloFusiello:


Dockerfile
```diff
@@ -3,41 +3,43 @@ FROM python:3.7.3-slim as base
 ENV AIRFLOW_HOME=/airflow
 ENV AIRFLOW__CORE__DAGS_FOLDER=${AIRFLOW_HOME}/dags
 ENV AIRFLOW__CORE__BASE_LOG_FOLDER=${AIRFLOW_HOME}/logs
-ENV AIRFLOW__CORE__EXECUTOR=LocalExecutor
+ENV AIRFLOW__CORE__EXECUTOR=CeleryExecutor
```
giorgiosironi: expected change (doesn't say anything about the persistence backend)

Dockerfile
```diff
@@ -3,41 +3,43 @@ FROM python:3.7.3-slim as base
...
 
-COPY ./requirements ./requirements
+COPY ./requirements.txt ./
 
 RUN pip install -U pip \
     && set -ex \
     && apt-get update -yq \
     && apt-get install -yq --no-install-recommends build-essential \
-    && pip install --no-cache-dir -r ./requirements/base.txt \
+    && pip install --no-cache-dir -r ./requirements.txt \
```
giorgiosironi: the simplification of having `dev` just being identical to the published image (eventually used in `prod`) is good for me (the direction of having development dependencies in `prod` wouldn't be)

docker-compose.yml
```diff
@@ -47,3 +49,11 @@ services:
   airflow_scheduler:
     <<: *airflow
     command: scheduler
+
+  airflow_worker:
+    <<: *airflow
+    command: worker
+    ports:
+      - 8793:8793
```
giorgiosironi: what's the port for?

GiancarloFusiello: This port is used by the web server to request logs from workers.

Dockerfile
GiancarloFusiello: I thought I would use a single requirements file until the need for dev/prod specific requirements emerges.

Dockerfile
```diff
@@ -3,41 +3,43 @@ FROM python:3.7.3-slim as base
...
 
 ENV PYTHONUNBUFFERED=true
 ENV PYTHONDONTWRITEBYTECODE=true
 
-RUN pip install --no-cache-dir -r ./requirements/dev.txt \
-    && rm -rf ~/.cache/* \
-    && rm -rf /tmp/*
+ENV AIRFLOW__CORE__EXPOSE_CONFIG=true
 
 ADD https://github.com/ufoscout/docker-compose-wait/releases/download/2.5.0/wait /wait
```
giorgiosironi: alternative is to run `wget` or `curl` inside a single `RUN` to retrieve this, but it's probably not available in the image and it makes little difference

dags/test_dag.py
```diff
@@ -0,0 +1,53 @@
...
+    print_this = python_operator.PythonOperator(task_id='print_this',
+                                                provide_context=True,
+                                                python_callable=print_something)
+
+    log_that = python_operator.PythonOperator(task_id='log_that',
+                                              provide_context=True,
+                                              python_callable=log_something)
+
+    # run tasks
+    bash_this >> print_this >> log_that
```
giorgiosironi: what a syntax :see_no_evil:

docker-compose.yml
giorgiosironi: so it may not be necessary to expose it to the host, unless something outside of the containers uses it?

dags/test_dag.py
```diff
@@ -0,0 +1,53 @@
+"""DAG running in response to a AWS S3 bucket change."""
```
giorgiosironi: we should run this as part of the builds in https://github.com/libero/elife-style-content-adapter-prototype/blob/master/.travis/smoke-test.sh
I'm not sure what is the waiting condition to understand it has finished. Perhaps move the `BashOperator` at the end and make it create a file inside a mounted volume, so that the host can check it appears?

docker-compose.yml
GiancarloFusiello: will try without exposing the port.

dags/test_dag.py
GiancarloFusiello: Having a look to see if there is something here we can leverage: https://airflow.apache.org/api.html

GiancarloFusiello: How about these? :
```
curl -X GET localhost:8080/api/experimental/test

{"status":"OK"}
```
After running `libero_test_dag`:
```
curl -X GET localhost:8080/api/experimental/dags/libero_test_dag/dag_runs

[{"dag_id":"libero_test_dag",
   "dag_run_url":"/admin/airflow/graph?dag_id=libero_test_dag&execution_date=2019-05-07+15%3A01%3A51%2B00%3A00",
   "execution_date":"2019-05-07T15:01:51+00:00",
   "id":1,
   "run_id":"manual__2019-05-07T15:01:51+00:00",
   "start_date":"2019-05-07T15:01:51.962440+00:00",
   "state":"success"}]
```

docker-compose.yml
GiancarloFusiello: doesn't need to be exposed. Removing.

dags/test_dag.py
giorgiosironi: :+1: it's a published API so very stable

GiancarloFusiello: I've added the status check for now. In the next PR I will add DAG tests.

GiancarloFusiello:
@giorgiosironi Ok to merge if all green?

PR #3, opened 2019-05-08
Create s3 buckets

GiancarloFusiello:
`awscli` service creates `incoming` and `expanded` s3 buckets and copies .zip file to `incoming` bucket when `docker-compose up` is executed.

.zip file contains JATS XML and a jpeg image referenced in the xml.


docker-compose.yml
```diff
@@ -1,7 +1,7 @@
 version: "3.4"
 
 services:
-  localstack-s3:
+  localstack:
```
giorgiosironi: I'm surprised there isn't a standard way to setup buckets into `localstack`, either through its internal CLI tools or via configuration variables

giorgiosironi: we do something similar to this PR in https://github.com/elifesciences/personalised-covers-formula/blob/a15351661fd2935dfd001bd679632e09601556b8/salt/personalised-covers/localstack.sls#L3

docker-compose.yml
```diff
@@ -1,7 +1,7 @@
...
     image: localstack/localstack:latest
     environment:
```
giorgiosironi: there's also `SERVICES=S3` but unsure what it does now:
https://github.com/elifesciences/personalised-covers-formula/blob/6c12a674b8e5f50e2f014340088d42d8007a3ef9/salt/pillar/personalised-covers.sls#L29-L31

Dockerfile.awscli
```diff
@@ -0,0 +1,10 @@
+FROM python:3.7.3-alpine3.9
```
giorgiosironi: https://hub.docker.com/r/mesosphere/aws-cli or any official image may help?

docker-compose.yml
GiancarloFusiello: This limits the services to start up. Updated.

Dockerfile.awscli
GiancarloFusiello: updated.

docker-compose.yml
GiancarloFusiello: I don't see anything other than using the aws cli.

scripts/create-s3-buckets.sh
```diff
@@ -0,0 +1,11 @@
...
+
+set -e
+
+# Executable that waits for a port at an address to be open.
+# For further details go to: https://github.com/ufoscout/docker-compose-wait
+/wait
+
+awslocal s3 mb s3://unstable-elife-style-content-adapter-incoming
+awslocal s3 mb s3://unstable-elife-style-content-adapter-expanded
+awslocal s3 cp ./elife-666-vor-r1.zip s3://unstable-elife-style-content-adapter-incoming
```
giorgiosironi: this archive is small enough - for anything more than 1MB https://git-lfs.github.com/ will be welcome (the hard limit is 100 MB per file on Github, but the soft limit for keeping the repository lean is much lower)

Dockerfile.awscli
```diff
@@ -0,0 +1,8 @@
+FROM imione/awscli-local
+
+ADD https://github.com/ufoscout/docker-compose-wait/releases/download/2.5.0/wait /wait
+RUN chmod +x /wait
+
+COPY ./scripts/create-s3-buckets.sh ./elife-666-vor-r1.zip ./
+
+CMD ["./create-s3-buckets.sh"]
```
giorgiosironi: this image build should be very fast now. You may want to move `CMD` to `command` in `docker-compose.yml` if you think the image will be useful to run other administration commands via `docker-compose run awscli ...`, so that there is no default command.

scripts/create-s3-buckets.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+set -e
+
+# Executable that waits for a port at an address to be open.
+# For further details go to: https://github.com/ufoscout/docker-compose-wait
+/wait
+
+awslocal s3 mb s3://unstable-elife-style-content-adapter-incoming
```
giorgiosironi: `unstable` is the name of an environment where this is deployed; the convention for the environment that is used locally is `dev`, or more in general `ci` if it is running in Travis CI or a similar place. I would go for `dev-elife-style-content-adapter-incoming`, similarly for the other bucket?

scripts/create-s3-buckets.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
```
giorgiosironi: there is nothing specific to `bash` in this script (even `set -e` seems portable) so this should be the best default. I usually run `shellcheck` on all scripts to check for this intricacies.

Dockerfile.awscli
GiancarloFusiello: updated.

PR #4, opened 2019-05-17
add s3 integrations to dags

GiancarloFusiello:
Adds code and configurations required to connect to s3 and test the current code base.


Dockerfile
```diff
@@ -1,35 +1,22 @@
...
-ENV AIRFLOW__CORE__EXECUTOR=CeleryExecutor
-ENV AIRFLOW__CORE__PLUGINS_FOLDER=${AIRFLOW_HOME}/plugins
-ENV AIRFLOW__CORE__CHILD_PROCESS_LOG_DIRECTORY=${AIRFLOW_HOME}/logs/scheduler
-
-ENV AIRFLOW__CORE__SQL_ALCHEMY_CONN=postgresql+psycopg2://postgres:example@db/airflow-db
-ENV AIRFLOW__CORE__LOAD_EXAMPLES=false
-ENV AIRFLOW__CORE__DAGS_ARE_PAUSED_AT_CREATION=false
-
-ENV AIRFLOW__CELERY__BROKER_URL=sqla+${AIRFLOW__CORE__SQL_ALCHEMY_CONN}
-ENV AIRFLOW__CELERY__RESULT_BACKEND=db+${AIRFLOW__CORE__SQL_ALCHEMY_CONN}
```
giorgiosironi: simply gone? Were they unused, or they have defaults?

GiancarloFusiello: Using configuration files instead.

dags/aws.py
```diff
@@ -0,0 +1,30 @@
...
+    """
+    returns a generator that lists keys/prefixes in an AWS S3 bucket.
+    """
+    client = get_aws_connection('s3')
+    paginator = client.get_paginator('list_objects_v2')
+    for page in paginator.paginate(**list_objects_v2_params):
+        if 'Contents' in page:
+            for item in page['Contents']:
+                yield item['Key']
+        elif 'CommonPrefixes' in page:
```
giorgiosironi: prefixes are akin to fake directories I think (if they end in `/`), what are they used for?

dags/process_elife_zip_dag.py
```diff
@@ -0,0 +1,107 @@
...
+        s3 = get_aws_connection('s3')
+        # store downloaded file in temp file
+        s3.download_fileobj(SOURCE_BUCKET, file_name, temp_file)
+        logger.debug('ZIPPED FILES= %s', ZipFile(temp_file).namelist())
+
+        folder_name = file_name.replace('.zip', '')
+        xml_path = None
+        for zipped_file_path in ZipFile(temp_file).namelist():
+            # skip mac os archive files
+            if '__MACOSX' not in zipped_file_path:
```
giorgiosironi: uh, should be extracted in a list of names to skip as hidden directories? I can think of similar examples from Windows or Linux

dags/process_elife_zip_dag.py
```diff
@@ -0,0 +1,107 @@
...
+                    # store zipped file in temp file
+                    inner_temp_file.write(ZipFile(temp_file).read(zipped_file_path))
+                    inner_temp_file.seek(0)
+                    s3_key = '%s/%s' % (folder_name, zipped_file_path)
+                    # upload stored zipped file preserving zip file structure
+                    s3.upload_fileobj(inner_temp_file, DESTINATION_BUCKET, s3_key)
+                    logger.info(
+                        '%s uploaded to %s/%s', zipped_file_path, DESTINATION_BUCKET, s3_key
+                    )
+                    if zipped_file_path.endswith('.xml'):
```
giorgiosironi: can there be more than 1 xml in the same zip? Not sure, but I would assert that there is only one so that an exception is thrown if there is a new case where >=2 are used

dags/process_elife_zip_dag.py
```diff
@@ -0,0 +1,107 @@
...
+    xml_path = context['task_instance'].xcom_pull(task_ids=previous_task)
+    logger.debug('FILE PASSED= %s', xml_path)
+    if xml_path is None:
+        raise ValueError('path to xml document was not passed from task %s' % previous_task)
+
+    # temporary files are securely stored on disk and automatically deleted when closed
+    with TemporaryFile() as temp_file:
+        s3 = get_aws_connection('s3')
+        # store downloaded file in temp file
+        s3.download_fileobj(DESTINATION_BUCKET, xml_path, temp_file)
```
giorgiosironi: the name of the function (`..._for_libero`) suggests some transformation, but it seems that it only downloads the file and parses it? The name could be simplified or, if the transformation is just an empty step at the moment, that step could be extracted for clarity into another function that we expect to grow later on.

dags/aws.py
GiancarloFusiello: Here I'm getting the top level "directory" as this should match the zip file name. I use this information to compare the incoming and expanded bucket to identify which zip files need to be worked on. This assumes that we want zipped files to be expanded in a directory matching the zip name.

dags/process_elife_zip_dag.py
GiancarloFusiello: Yes, I wasn't sure how far to go with this. Either have a list of patterns to skip or do not skip anything. What do you think?

dags/process_elife_zip_dag.py
GiancarloFusiello: Yes I also made an assumption here that there would only be one xml file. I'm unsure if there can be multiple xml files. Do you know if `manifest.xml` files are included sometimes?

dags/process_elife_zip_dag.py
giorgiosironi: I think existing projects take a whitelist approach and have a list of patterns they look for (XML, images, and so on) ignoring all the rest. However, that's more for backward compatibility in case new types of files are added (e.g. videos or supplementary files like a dataset in .csv format); if a content processor sent .zip archives that contained empty or unused folders I would send them back as incorrect input.

dags/process_elife_zip_dag.py
GiancarloFusiello: correct. I only added the s3 functionality at this point. I'll add the xml transformation in the next PR as this one is quite large.

dags/process_elife_zip_dag.py
GiancarloFusiello: What would you say is the requirement here?

dags/process_elife_zip_dag.py
giorgiosironi: As far as I know there are no manifests at this stage (I've seen them in MECA, a format used by xpub but not here.)

giorgiosironi: Good to assume that there is only one file, would assert that this is the case.

dags/process_elife_zip_dag.py
giorgiosironi: Path of least resistance: take the files you need and process them; if additional, unnecessary files are making the code fail, let it fail.

GiancarloFusiello: added a list of accepted file extensions.

dags/process_elife_zip_dag.py
GiancarloFusiello: check added.

dags/process_elife_zip_dag.py
```diff
@@ -0,0 +1,118 @@
...
+    'start_date': timezone.utcnow(),
+    'email': ['libero@example.com'],
+    'email_on_failure': False,
+    'email_on_retry': False,
+    'retries': 3,
+    'retry_delay': timedelta(seconds=5)
+}
+
+
+def extract_zipped_files_to_bucket(**context):
```
giorgiosironi: This module is specific for the .zip format, but I wonder if there is a more precise name we can use (article archive? article delivery?) to avoid the zip details leaking over code that is not tied to that. At the moment however, "zip" is practically not mentioned outside of this function, so no problem.

dags/process_elife_zip_dag.py
```diff
@@ -0,0 +1,118 @@
...
+                  (len(xml_files), file_name, xml_files))
+        assert len(xml_files) == 1, message
+
+        folder_name = file_name.replace('.zip', '')
+        xml_path = None
+        for zipped_file_path in ZipFile(temp_file).namelist():
+            if Path(zipped_file_path).suffix in ALLOWED_EXTENSIONS:
+                with TemporaryFile() as inner_temp_file:
+                    # store zipped file in temp file
+                    inner_temp_file.write(ZipFile(temp_file).read(zipped_file_path))
```
giorgiosironi: tried using `Zipfile.open()`? It may return a file-like object that you can directly pass to the `s3` client

dags/trigger_dag.py
```diff
@@ -0,0 +1,85 @@
+"""
+DAG identify zip files to process from s3 bucket and trigger dag for each zip file
```
giorgiosironi: (the Python module) could be a more generic name e.g. `watch_dag` or `s3_watch_dag`, but it's more complicated than this

dags/trigger_dag.py
```diff
@@ -0,0 +1,85 @@
...
+def run_dag_for_each_file(**context):
+    file_names = get_return_value_from_previous_task(**context)
+    dag_to_trigger = process_elife_zip_dag.dag.dag_id
+    for file_name in file_names:
+        trigger_dag(dag_id=dag_to_trigger,
+                    run_id='{}_{}'.format(file_name, uuid4()),
+                    conf=json.dumps({'file': file_name}),
+                    execution_date=None,
+                    replace_microseconds=False)
+    logger.debug('triggered %s for %s files: %s' % (dag_to_trigger, len(file_names), file_names))
```
giorgiosironi: raising this log level?

giorgiosironi: (if `file_names` is non empty maybe)

tests/conftest.py
```diff
@@ -0,0 +1,48 @@
...
+from tests.factories import DAGFactory, PythonOperatorFactory
+
+
+def pytest_runtest_setup():
+
+    def block_lookup(*args, **kwargs):
+        raise SocketBlockedError
+
+    disable_socket()
+    socket.getaddrinfo = block_lookup
```
giorgiosironi: if this comes from Boto idiosyncrasies, state it in the docstring of this function

tests/mocks.py
```diff
@@ -0,0 +1,116 @@
+import datetime
+
+from dateutil.tz import tzlocal
+
+from tests.assets import get_asset
+
+
+class s3ClientMock:
```
giorgiosironi: for future expansion, follow the Only Mock Types You Own principle and introduce an abstraction layer over Boto if `S3Client` gets used by many different tasks

tests/test_aws.py
```diff
@@ -0,0 +1,50 @@
...
+
+@pytest.mark.parametrize('url', [None, 'http://test-url.com'])
+def test_get_aws_connection(url):
+    import dags.aws
+    dags.aws.ENDPOINT_URL = url
+
+    conn = get_aws_connection('s3')
+
+    # boto3 client should use the default address if an endpoint is not specified
+    if not url:
```
giorgiosironi: consider two separate tests

tests/test_process_elife_zip_dag.py
```diff
@@ -0,0 +1,85 @@
...
+        assert str(error.value) == msg
+
+
+def test_extract_zipped_files_to_bucket_raises_exception_if_no_xml_files_in_zip(mocker, context):
+    mocker.patch('boto3.client', new_callable=s3ClientMock)
+    file_name = 'elife-666-vor-r1.zip'
+    xml_files = []
+    context['dag_run'].conf = {'file': file_name}
+    namelist = mocker.patch('zipfile.ZipFile.namelist')
+    namelist.return_value = xml_files
```
giorgiosironi: consider extracting an helper function if there is duplicated code between some tests

tests/test_trigger_dag.py
```diff
@@ -0,0 +1,45 @@
...
+    # populate expected return value of previous task
+    ti = context['dag_run'].get_task_instances()[0]
+    ti.xcom_push(key='return_value', value=file_names)
+
+    trigger = mocker.patch('dags.trigger_dag.trigger_dag')
+
+    run_dag_for_each_file(**context)
+
+    assert trigger.call_count ==  expected
+    if trigger.called:
```
giorgiosironi: can we assert anything in the `else` case?

dags/process_elife_zip_dag.py
GiancarloFusiello: renamed to `extract_archived_files_to_bucket`

dags/process_elife_zip_dag.py
GiancarloFusiello: I had a look into this and s3 doesn't support streamed uploads at this time. There are many options for uploading files but all rely on a file object of some sort.

dags/trigger_dag.py
GiancarloFusiello: updated logging level and added assertion to check file_names type.

tests/conftest.py
GiancarloFusiello: description added.

tests/mocks.py
GiancarloFusiello: refactored which removes the need for mocked responses.

tests/test_aws.py
GiancarloFusiello: refactored

tests/test_process_elife_zip_dag.py
GiancarloFusiello: refactored into a single test.

tests/test_trigger_dag.py
GiancarloFusiello: I couldn't think of anything. The test for an empty set is to ensure the trigger is not called.

PR #5, opened 2019-06-03
add s3 streaming objects

GiancarloFusiello:
replace the use of temp files with in memory objects for individual files/s3 objects

### DIFF
--- a/dags/process_elife_zip_dag.py
+++ b/dags/process_elife_zip_dag.py
@@ -63,18 +63,20 @@ def extract_archived_files_to_bucket(**context):
         xml_path = None
         for zipped_file_path in ZipFile(temp_file).namelist():
             if Path(zipped_file_path).suffix in ALLOWED_EXTENSIONS:
-                with TemporaryFile() as inner_temp_file:
-                    # store zipped file in temp file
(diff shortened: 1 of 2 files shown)


dags/process_elife_zip_dag.py
```diff
@@ -63,18 +63,20 @@ def extract_archived_files_to_bucket(**context):
...
-                    logger.info(
-                        '%s uploaded to %s/%s', zipped_file_path, DESTINATION_BUCKET, s3_key
-                    )
-                    if zipped_file_path.endswith('.xml'):
-                        xml_path = s3_key
+                s3_key = '%s/%s' % (folder_name, zipped_file_path)
+                s3.put_object(
+                    Bucket=DESTINATION_BUCKET,
+                    Key=s3_key,
+                    Body=ZipFile(temp_file).read(zipped_file_path)
```
giorgiosironi: should `ZipFile(temp_file)` be extracted before the loop? Should only need to create one object and issue many `read()` call towards it?

GiancarloFusiello: The contents are never extracted. `ZipFile` can read the contents of the archive without extracting the files.

PR #6, opened 2019-06-06
add xml transformations

GiancarloFusiello:
make changes according to https://github.com/libero/libero/issues/155

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,6 +20,7 @@ COPY ./dags ${AIRFLOW_HOME}/dags
 
 RUN chown -R airflow: .
 
+
 FROM base as dev
 
 ENV PYTHONUNBUFFERED=true
--- a/config/airflow.dev.cfg
+++ b/config/airflow.dev.cfg
@@ -1,8 +1,9 @@
 [libero]
 # s3
-endpoint_url = http://localstack:4572
+boto_endpoint_url = http://localstack:4572
 source_bucket = dev-elife-style-content-adapter-incoming
 destination_bucket = dev-elife-style-content-adapter-expanded
+base_url = https://%(destination_bucket)s.s3.amazon.aws.com
 service_name = content-store
 service_url =
 
--- a/config/airflow.dev.cfg
+++ b/config/airflow.dev.cfg
@@ -3,7 +3,7 @@
 boto_endpoint_url = http://localstack:4572
 source_bucket = dev-elife-style-content-adapter-incoming
 destination_bucket = dev-elife-style-content-adapter-expanded
-base_url = https://%(destination_bucket)s.s3.amazon.aws.com
+base_url = https://%(destination_bucket)s.s3.amazonaws.com
 service_name = content-store
 service_url =
 
--- a/config/airflow.dev.cfg
+++ b/config/airflow.dev.cfg
@@ -3,6 +3,8 @@
 endpoint_url = http://localstack:4572
 source_bucket = dev-elife-style-content-adapter-incoming
 destination_bucket = dev-elife-style-content-adapter-expanded
+service_name = content-store
+service_url =
 
 [core]
 # The folder where your airflow pipelines live, most likely a
--- a/dags/process_elife_zip_dag.py
+++ b/dags/process_elife_zip_dag.py
@@ -21,7 +21,7 @@ from task_helpers import get_previous_task_name
 
 ALLOWED_EXTENSIONS = {'.jpg', '.jpeg', '.pdf', '.xml'}
 
-ENDPOINT_URL = configuration.conf.get('libero', 'endpoint_url') or 'https://s3.amazonaws.com'
+BASE_URL = configuration.conf.get('libero', 'base_url')
 SOURCE_BUCKET = configuration.conf.get('libero', 'source_bucket')
 DESTINATION_BUCKET = configuration.conf.get('libero', 'destination_bucket')
 SERVICE_NAME = configuration.conf.get('libero', 'service_name')
@@ -112,7 +112,7 @@ def prepare_jats_xml_for_libero(**context):
     root = article_xml.getroot()
     root.set(
         '{%s}base' % XML_NAMESPACE,
-        '%s/%s/%s' % (ENDPOINT_URL, DESTINATION_BUCKET, key)
+        '%s/%s' % (BASE_URL, key)
     )
 
     # create libero xml
--- a/dags/process_elife_zip_dag.py
+++ b/dags/process_elife_zip_dag.py
@@ -7,20 +7,25 @@ from datetime import timedelta
 from io import BytesIO
 from pathlib import Path
 from tempfile import TemporaryFile
+from xml.dom import XML_NAMESPACE
 from zipfile import ZipFile
 
 from airflow import DAG, configuration
 from airflow.operators import python_operator
 from airflow.utils import timezone
 from lxml import etree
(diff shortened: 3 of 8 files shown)


Dockerfile
```diff
@@ -20,6 +20,7 @@ COPY ./dags ${AIRFLOW_HOME}/dags
 
 RUN chown -R airflow: .
 
+
```
giorgiosironi: intentional?

dags/process_elife_zip_dag.py
```diff
@@ -90,8 +95,38 @@ def prepare_jats_xml_for_libero(**context):
 
     s3 = get_aws_connection('s3')
     response = s3.get_object(Bucket=DESTINATION_BUCKET, Key=xml_path)
-    xml = etree.parse(BytesIO(response['Body'].read()))
-    return etree.tostring(xml)
+    article_xml = etree.parse(BytesIO(response['Body'].read()))
+
+    # get article id
+    article_id = article_xml.xpath(
+        '/article/front/article-meta/article-id[@pub-id-type="publisher-id"]'
```
giorgiosironi: the XPath queries may benefit from being extracted as string constants, I do this with [CSS queries](https://github.com/elifesciences/elife-spectrum/blob/master/spectrum/xpub.py#L25-L27) for example

Dockerfile
GiancarloFusiello: yes. adding another line makes it slightly easier to read (in my opinion)

dags/process_elife_zip_dag.py
GiancarloFusiello: I would agree if the paths were used multiple places but at the moment it's only for single use.

dags/process_elife_zip_dag.py
```diff
@@ -90,8 +95,38 @@ def prepare_jats_xml_for_libero(**context):
...
+    )[0].text
+
+    # add jats prefix to jats tags
+    for element in article_xml.iter():
+        if not element.prefix:
+            element.tag = '{http://jats.nlm.nih.gov}%s' % element.tag
```
giorgiosironi: I thought the input, being JATS, to have a correctly defined namespace for the JATS elements; probably at the top level, perhaps as a default namespace. However this doesn't seem to be the case?
```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE article PUBLIC "-//NLM//DTD JATS (Z39.96) Journal Archiving and Interchange DTD v1.1d3 20150301//EN" "JATS-archivearticle1.dtd">
<article xmlns:xlink="http://www.w3.org/1999/xlink" article-type="article-commentary" dtd-version="1.1d3">
<front><journal-meta>
...
```

GiancarloFusiello: Correct. I thought this was the case as well but it seems that the declaration of a name is not enough for lxml to actually add the namespace to elements.

config/airflow.dev.cfg
```diff
@@ -3,6 +3,8 @@
 endpoint_url = http://localstack:4572
 source_bucket = dev-elife-style-content-adapter-incoming
 destination_bucket = dev-elife-style-content-adapter-expanded
+service_name = content-store
+service_url =
```
giorgiosironi: would expect this to be needed shortly, not in this PR however as it was not talking to that service. Can be left in, little harm

GiancarloFusiello: yes, in an ideal world I would have a container that either mimics the content-store or can receive a put request. For now I'm just mocking the value in tests but the key name still needs to be present in the config otherwise an exception will be raised.

dags/process_elife_zip_dag.py
giorgiosironi: What I meant was I was expecting a `xmlns` attribute in the root element of the input XML. Turns out it's not there, so :man_shrugging: no need to change anything.

GiancarloFusiello: oh yeah...:man_shrugging:

dags/process_elife_zip_dag.py
```diff
@@ -7,20 +7,25 @@
...
 from lxml import etree
+from lxml.builder import ElementMaker
 
-from aws import get_aws_connection, list_bucket_keys_iter
+from aws import get_aws_connection
 from task_helpers import get_previous_task_name
 
 ALLOWED_EXTENSIONS = {'.jpg', '.jpeg', '.pdf', '.xml'}
 
+ENDPOINT_URL = configuration.conf.get('libero', 'endpoint_url') or 'https://s3.amazonaws.com'
```
giorgiosironi: I think the naming of this configuration setting can be improved as `endpoint` in AWS jargon is typically the target of an API call done by the SDK (which follows a XML-based format with signatures, etc.)
What we are after here is a more abstract URL prefix where the images (or other assets) will be accessible. In theory, you could even put a proxy like a gateway that authenticates or caches the files in between. I would configure the whole `EXPANDED_URL` including the bucket name.

giorgiosironi: Moreover, I just discovered to be [future proof](https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/) buckets should be accessed via a `https://${bucket-name}.s3.amazonaws.com` URL.

GiancarloFusiello: This is used here: https://github.com/libero/elife-style-content-adapter-prototype/blob/master/dags/aws.py#L7

I could rename it to `boto_endpoint_url`?

dags/process_elife_zip_dag.py
```diff
@@ -90,8 +95,38 @@ def prepare_jats_xml_for_libero(**context):
...
+    for element in article_xml.iter():
+        if not element.prefix:
+            element.tag = '{http://jats.nlm.nih.gov}%s' % element.tag
+
+    # add xml:base attribute to article element
+    key = Path(xml_path).parent.stem
+    root = article_xml.getroot()
+    root.set(
+        '{%s}base' % XML_NAMESPACE,
+        '%s/%s/%s' % (ENDPOINT_URL, DESTINATION_BUCKET, key)
```
giorgiosironi: Here's where it gets complicated. After expansion, the files are in the expanded bucket which at the moment is configured to be publicly accessible so that a `content-store` can pull the files without authentication. It's difficult to add authentication without being too much AWS-specific, as the `content-store` would have to generate a signature to add to requests in order to see the files.
So at the moment the files will be publicly accessible as soon as you get here, if you know the file names. This is just for seconds in this workflow, as than the `content-store` makes them public anyway not having yet a concept of `preview` article not ready for publication. Medium-term will need to do something about it, could be as simple as expanding the files into an unguessable folder:
```
https://unstable-expanded.s3.amazon.aws.com/fdb8d906-77cb-470d-9090-40a046e81bf9/elife-666-fig1.jpg
```
We probably don't want to implement this yet, just a possibility (of many).

config/airflow.dev.cfg
```diff
@@ -1,8 +1,11 @@
 [libero]
 # s3
-endpoint_url = http://localstack:4572
+boto_endpoint_url = http://localstack:4572
 source_bucket = dev-elife-style-content-adapter-incoming
 destination_bucket = dev-elife-style-content-adapter-expanded
+base_url = https://%(destination_bucket)s.s3.amazon.aws.com
```
giorgiosironi: it's `amazonaws`, single word

PR #7, opened 2019-06-07
send put request to content store

GiancarloFusiello:
As per https://github.com/libero/libero/issues/156 Adds a new task to `process_elife_zip_dag` to send a PUT request to the specified service (content store)

### DIFF
--- a/dags/process_elife_zip_dag.py
+++ b/dags/process_elife_zip_dag.py
@@ -129,7 +129,7 @@ def wrap_article_in_libero_xml_and_send_to_service(**context):
 
     # make PUT request to service
     response = requests.put(
-        SERVICE_URL,
+        '%s/items/%s/version/1' % (SERVICE_URL, article_id),
         data=etree.tostring(xml, xml_declaration=True, encoding='UTF-8'),
         headers={'content-type': 'application/xml'}
     )
--- a/dags/process_elife_zip_dag.py
+++ b/dags/process_elife_zip_dag.py
@@ -129,7 +129,7 @@ def wrap_article_in_libero_xml_and_send_to_service(**context):
 
     # make PUT request to service
     response = requests.put(
-        '%s/items/%s/version/1' % (SERVICE_URL, article_id),
+        '%s/items/%s/versions/1' % (SERVICE_URL, article_id),
         data=etree.tostring(xml, xml_declaration=True, encoding='UTF-8'),
         headers={'content-type': 'application/xml'}
     )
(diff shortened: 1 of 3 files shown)


dags/process_elife_zip_dag.py
```diff
@@ -94,6 +96,19 @@ def prepare_jats_xml_for_libero(**context):
     return etree.tostring(xml)
 
 
+def send_put_request_to_service(**context):
+    return_value = get_return_value_from_previous_task(**context)
```
giorgiosironi: IIRC this means the XML is stored (long-term) in Airflow's database?

dags/process_elife_zip_dag.py
```diff
@@ -94,6 +96,19 @@ def prepare_jats_xml_for_libero(**context):
...
+    message = 'the previous task did not return data to send to %s' % SERVICE_URL
+    assert return_value is not None, message
+    response = requests.put(
+        SERVICE_URL,
```
giorgiosironi: this being a PUT request, it is a dynamic URL containing the type and id of the item

dags/process_elife_zip_dag.py
GiancarloFusiello: Correct, return values are stored. For how long? I'm not sure but this is part of Airflow's model.

dags/process_elife_zip_dag.py
giorgiosironi: In the unstable environment, this will be http://unstable--api-gateway.libero.pub:8081/scholarly-articles/items/article1/versions/1 so the prefix `http://unstable--api-gateway.libero.pub:8081/scholarly-articles/` should be made configurable.
`article1` is the id of the article, extracted from the XML I guess.

GiancarloFusiello: The URL was missing from the task/Issue. What is the structure of the url?

GiancarloFusiello: This URL indicates that I need to extract the article ID and revision in order form the URL. Do you think I should move making the request to the previous task?

dags/process_elife_zip_dag.py
giorgiosironi: Theoretically you could have tasks running on different workers, so there isn't a shared environment between them where to store data/files apart from S3. So it is conveniente for Airflow to pass the data around. Just curious on whether it's possible to eventually prune the data for old workflows from the database to free space.

GiancarloFusiello: This is the only documentation I can find about it but I'm going through the XCOM source could reveal its inner workings: https://airflow.apache.org/concepts.html#xcoms

PR #12, opened 2019-06-10
add local content store

GiancarloFusiello:


content-store.py
```diff
@@ -0,0 +1,12 @@
...
+
+flask_app = flask.Flask(__name__)
+
+@flask_app.route('/items/<articleid>/versions/1', methods=['PUT'])
+def items(articleid):
+    return flask.Response(flask.request.data, mimetype="application/xml")
+
+
+if __name__ == '__main__':
+    flask_app.run(host='0.0.0.0', port=8090)
```
giorgiosironi: this runs in a container with its own port space, so it should be no harm to make it the standard port 80?

docker-compose.yml
```diff
@@ -81,3 +81,13 @@ services:
   airflow_worker:
     <<: *airflow
     command: worker
+
+  content-store:
+    image: jats-ingester_airflow_webserver
```
giorgiosironi: this `image` shouldn't be needed `(`build` is used instead) so can be removed?

GiancarloFusiello: image is used and if the image is not found build is used as a fallback: https://docs.docker.com/compose/compose-file/#image

content-store.py
GiancarloFusiello: updated.

PR #16, opened 2019-06-13
add trailing slash to xml:base uri

GiancarloFusiello:
In reference to https://github.com/libero/libero/issues/173

### DIFF
--- a/tests/test_process_elife_zip_dag.py
+++ b/tests/test_process_elife_zip_dag.py
@@ -1,5 +1,6 @@
 from io import BytesIO
 from pathlib import Path
+from xml.dom import XML_NAMESPACE
 
 import pytest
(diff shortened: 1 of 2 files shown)


tests/test_process_elife_zip_dag.py
```diff
@@ -92,6 +93,7 @@ def test_wrap_article_in_libero_xml_and_send_to_service(context, s3_client, requ
     article = xml.xpath('//libero:item/jats:article', namespaces=namespaces)[0]
     assert article is not None
     assert len(article.getchildren()) > 0
+    assert article.attrib['{%s}base' % XML_NAMESPACE].endswith('/')
```
giorgiosironi: I can't think of a more abstract assertion to use right now :thinking:

PR #18, opened 2019-06-20
Convert images to jpeg. Update image references in article xml

GiancarloFusiello:
In reference to: https://github.com/libero/libero/issues/182

### DIFF
--- a/dags/process_elife_zip_dag.py
+++ b/dags/process_elife_zip_dag.py
@@ -17,11 +17,13 @@ from airflow.operators import python_operator
 from airflow.utils import timezone
 from lxml import etree
 from lxml.builder import ElementMaker
+from PIL import Image
 
 from aws import get_aws_connection
 from task_helpers import get_previous_task_name
 
-ALLOWED_EXTENSIONS = {'.jpg', '.jpeg', '.pdf', '.xml'}
+ALLOWED_EXTENSIONS = {'.jpg', '.jpeg', '.pdf', '.xml'}  # only upload these file types
+IMAGE_EXTENSIONS = {'.jpg', '.jpeg', '.png', '.tif', '.tiff'}
 
 BASE_URL = configuration.conf.get('libero', 'base_url')
 SOURCE_BUCKET = configuration.conf.get('libero', 'source_bucket')
@@ -41,6 +43,14 @@ default_args = {
     'retries': 3,
     'retry_delay': timedelta(seconds=5)
 }
+
+
+def is_allowed(file_name: str) -> bool:
+    return Path(file_name).suffix in ALLOWED_EXTENSIONS
+
+
+def is_image(file_name: str) -> bool:
+    return Path(file_name).suffix in IMAGE_EXTENSIONS
 
 
 def extract_archived_files_to_bucket(**context):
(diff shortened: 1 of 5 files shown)


dags/process_elife_zip_dag.py
```diff
@@ -69,17 +79,38 @@ def extract_archived_files_to_bucket(**context):
...
 
         # extract zip to cloud bucket
-        folder_name = zip_file_name.replace('.zip', '')
+        folder_name = zip_file_name.rstrip('.zip')
+        file_to_upload = None
+
         for zipped_file_path in ZipFile(temp_file).namelist():
-            if Path(zipped_file_path).suffix in ALLOWED_EXTENSIONS:
+
+            if is_image(zipped_file_path) and not is_allowed(zipped_file_path):
```
giorgiosironi: time for some exploratory testing: take the TIFF files from the various articles https://github.com/elifesciences/elife-spectrum/tree/master/spectrum/templates and pass them through this conversion to see whether anything explodes. This would be a small sample, then we can take a larger one from the `prod-elife-published` bucket.

dags/process_elife_zip_dag.py
```diff
@@ -17,11 +17,13 @@
...
 from lxml import etree
 from lxml.builder import ElementMaker
+from PIL import Image
 
 from aws import get_aws_connection
 from task_helpers import get_previous_task_name
 
-ALLOWED_EXTENSIONS = {'.jpg', '.jpeg', '.pdf', '.xml'}
+ALLOWED_EXTENSIONS = {'.jpg', '.jpeg', '.pdf', '.xml'}  # only upload these file types
+IMAGE_EXTENSIONS = {'.jpg', '.jpeg', '.png', '.tif', '.tiff'}
```
giorgiosironi: seems reasonable so far as source images, in the end eLife uses only `.tif` for example

dags/process_elife_zip_dag.py
GiancarloFusiello: In the assets folder there is `elife-36842-vor-r3.zip` which contains many tiff files. I think the tests you're thinking of we can do afterward as part of adding demo content to unstable. If there's a bug I'll raise a new issue.

giorgiosironi: I guess https://github.com/libero/jats-ingester/pull/18/files#diff-5af835e7fbb86db53030f149d2de8d31R22 is the main coverage for this code then (separating code coverage from input coverage, which is the postponed part).

dags/process_elife_zip_dag.py
```diff
@@ -69,17 +79,38 @@ def extract_archived_files_to_bucket(**context):
...
         for zipped_file_path in ZipFile(temp_file).namelist():
-            if Path(zipped_file_path).suffix in ALLOWED_EXTENSIONS:
+
+            if is_image(zipped_file_path) and not is_allowed(zipped_file_path):
+                # convert image to jpeg
+                file_to_upload = BytesIO()
+                image = Image.open(BytesIO(ZipFile(temp_file).read(zipped_file_path)))
+                try:
+                    image.save(file_to_upload, 'JPEG')
+                except IOError:
```
giorgiosironi: anything more specific to catch? A disk space exhausted or permission error would try a new conversion, but there is only an error on color spaces here that we are trying to work around?

dags/process_elife_zip_dag.py
```diff
@@ -109,14 +140,20 @@ def wrap_article_in_libero_xml_and_send_to_service(**context):
         '/article/front/article-meta/article-id[@pub-id-type="publisher-id"]'
     )[0].text
 
+    # update image references to .jpg
```
giorgiosironi: looks like a (possibly) temporary change, maybe put it into a separate function so that it's easy to understand what to remove later on?

dags/process_elife_zip_dag.py
GiancarloFusiello: tiff images have an alpha channel that is not supported in jpeg images (as far as I understand). So instead of checking the file type (of which there could be many) this is an alternative check. In most cases PIL can convert images without having to do any conversions, meaning `image.save` should work most of the time.

dags/process_elife_zip_dag.py
GiancarloFusiello: Surely it's easier to remove this code block rather than removing a function call then having to find the function, which is somewhere else, and remove that also?

dags/process_elife_zip_dag.py
GiancarloFusiello: Also, these changes are made in memory.

dags/process_elife_zip_dag.py
giorgiosironi: As long as it seems isolated from the rest :+1:

giorgiosironi: (for me it's usually easier to remove a call to a function because it doesn't share the same scope as the calling code, e.g. I don't have to check that variables defined in the function aren't referenced outside of it)

GiancarloFusiello: ok. I see you point. In this instance, this chunk of code is updating the xml used in the rest of the function so it's isolated.

PR #19, opened 2019-06-21
Remove upload restrictions

GiancarloFusiello:
In reference to: https://github.com/libero/libero/issues/183


GiancarloFusiello:
draft until https://github.com/libero/jats-ingester/pull/18 is merged

giorgiosironi:
Needs a rebase/merge of `master`?

GiancarloFusiello:
> Needs a rebase/merge of `master`?

done.

dags/process_elife_zip_dag.py
```diff
@@ -45,12 +42,12 @@
...
 
-def is_allowed(file_name: str) -> bool:
-    return Path(file_name).suffix in ALLOWED_EXTENSIONS
-
-
-def is_image(file_name: str) -> bool:
-    return Path(file_name).suffix in IMAGE_EXTENSIONS
+def is_image(file: bytes) -> bool:
+    try:
+        Image.open(BytesIO(file))
```
giorgiosironi: from [the documentation](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.open) it seems this method is lazy enough not to read the contents of the files in memory, but it states that there is a file descriptor being left open for later usage. Does it need a `close()` call then?

giorgiosironi: Oh wait, the bytes are already in memory... so should they? If we are only identifying whether the file is an image or not we shouldn't need to read all of it right? e.g. It could be a 4GB video

giorgiosironi: I think this makes sense from the context of a caller but less in a generic `is_image` function :thinking:

dags/process_elife_zip_dag.py
```diff
@@ -80,37 +77,35 @@ def extract_archived_files_to_bucket(**context):
...
-                    Key=s3_key,
-                    Body=file_to_upload
-                )
-                logger.info(
-                    '%s uploaded to %s/%s',
-                    zipped_file_path,
-                    DESTINATION_BUCKET,
-                    folder_name
-                )
+            s3_key = '%s/%s' % (folder_name, zipped_file_path)
```
giorgiosironi: There is a modularity argument for separating the upload of the archive files to the bucket from the additional conversion of any file there to another format. It does imply re-downloading the TIF images, but allows the conversion to run on different workers/containers/nodes. Arguments for and against this separation?

GiancarloFusiello: How would you know which files to download?

giorgiosironi: Resolved in person: file extension is all that's needed for now.

config/airflow.dev.cfg
```diff
@@ -6,6 +6,7 @@ destination_bucket = dev-jats-ingester-expanded
 base_url = https://%(destination_bucket)s.s3.amazonaws.com
 service_name = content-store
 service_url = http://content-store
+temp_directory =
```
giorgiosironi: we probably won't even mount a volume for this directory as a start, but will configure the `/tmp` directory inside the container

dags/process_elife_zip_dag.py
```diff
@@ -19,17 +19,15 @@
...
-IMAGE_EXTENSIONS = {'.jpg', '.jpeg', '.png', '.tif', '.tiff'}
+from aws import get_aws_connection, list_bucket_keys_iter
+from task_helpers import get_file_name_passed_to_dag_run_conf_file
 
 BASE_URL = configuration.conf.get('libero', 'base_url')
 SOURCE_BUCKET = configuration.conf.get('libero', 'source_bucket')
 DESTINATION_BUCKET = configuration.conf.get('libero', 'destination_bucket')
 SERVICE_NAME = configuration.conf.get('libero', 'service_name')
 SERVICE_URL = configuration.conf.get('libero', 'service_url')
+TEMP_DIRECTORY = configuration.conf.get('libero', 'temp_directory') or None
```
giorgiosironi: `None` means the Python standard library will choose where, I guess

dags/process_elife_zip_dag.py
```diff
@@ -45,65 +43,43 @@
...
-                    image.save(file_to_upload, 'JPEG')
-                file_to_upload = file_to_upload.getvalue()
-                zipped_file_path = re.sub(r'\.\w+$', '.jpg', zipped_file_path)
-
-            if is_allowed(zipped_file_path):
+
+        # extract zip to cloud bucket
+        for zipped_file_path in ZipFile(temp_zip_file).namelist():
+            with TemporaryFile(dir=TEMP_DIRECTORY) as temp_zipped_file:
+                temp_zipped_file.write(ZipFile(temp_zip_file).read(zipped_file_path))
```
giorgiosironi: strange variable name: the file being written is not zipped anymore, `temp_expanded_file`|`temp_extracted_file`|`temp_unzipped_file`?

dags/process_elife_zip_dag.py
```diff
@@ -113,47 +89,90 @@ def extract_archived_files_to_bucket(**context):
...
-def wrap_article_in_libero_xml_and_send_to_service(**context):
-    # get xml path passed from previous task
-    previous_task = get_previous_task_name(**context)
-    xml_path = context['task_instance'].xcom_pull(task_ids=previous_task)
-    logger.info('XML PATH PASSED FROM PREVIOUS TASK= %s', xml_path)
-    message = 'path to xml document was not passed from task %s' % previous_task
-    assert xml_path is not None, message
+def convert_tiff_images_in_expanded_bucket_to_jpeg_images(**context):
+    zip_file_name = get_file_name_passed_to_dag_run_conf_file(**context)
+    for key in list_bucket_keys_iter(Bucket=DESTINATION_BUCKET, Prefix=zip_file_name.rstrip('.zip')):
```
giorgiosironi: I think  `Prefix` should add a trailing slash? e.g. `elife-123` should not list keys like `elife-1234/elife-1234-fig1.tif`

giorgiosironi: Also, [`rstrip`](https://docs.python.org/3/library/stdtypes.html#str.rstrip) doesn't do exactly what you want here:
```
>>> "hawaian.pizz".rstrip(".zip")
'hawaian'
```

dags/process_elife_zip_dag.py
```diff
@@ -113,47 +89,90 @@ def extract_archived_files_to_bucket(**context):
...
+        if key.endswith('.tif'):
+            with TemporaryFile(dir=TEMP_DIRECTORY) as temp_tiff_file:
+                s3 = get_aws_connection('s3')
```
giorgiosironi: can `s3` be pulled up and created before the loops? There will be no performance problem, but it's a service object with a longer lifecycle, could be mocked in tests, etc.

dags/process_elife_zip_dag.py
```diff
@@ -113,47 +89,90 @@ def extract_archived_files_to_bucket(**context):
...
+                s3.download_fileobj(
+                    Bucket=DESTINATION_BUCKET,
+                    Key=key,
+                    Fileobj=temp_tiff_file
+                )
+                image = Image.open(temp_tiff_file)
+                output = image.convert('RGB')
```
giorgiosironi: I recall this is to remove an alpha channel, is there a good link to include as a comment as a reference for this operation? Not an expert but input TIFFs could have all sorts of color spaces as far as I know, especially depending on whether they were also meant for print or not.

giorgiosironi: e.g. a Pillow documentation page or cookbook describing how to convert TIFFs

dags/process_elife_zip_dag.py
```diff
@@ -113,47 +89,90 @@ def extract_archived_files_to_bucket(**context):
...
+                temp_jpeg = BytesIO()
+                output.save(temp_jpeg, 'JPEG')
+                key = re.sub(r'\.\w+$', '.jpg', key)
```
giorgiosironi: I like the generic match in the regex, allowing other extensions to be introduced easily

dags/process_elife_zip_dag.py
```diff
@@ -113,47 +89,90 @@ def extract_archived_files_to_bucket(**context):
...
+                s3.put_object(
+                    Bucket=DESTINATION_BUCKET,
+                    Key=key,
+                    Body=temp_jpeg.getvalue()
+                )
+                logger.info('%s uploaded to %s',key, DESTINATION_BUCKET)
+
 
+def update_tiff_references_to_jpeg_in_article(**context):
```
giorgiosironi: from the name it seems like a good decomposition along task boundaries: I can imagine a real journal rolling out the conversion step first, and then adding this XML update later when the output images have been verified to work.

dags/process_elife_zip_dag.py
```diff
@@ -113,47 +89,90 @@ def extract_archived_files_to_bucket(**context):
...
+    for element in article_xml.xpath(
+            '//*[@mimetype="image" and @mime-subtype="tiff"]'):
+        href = '{http://www.w3.org/1999/xlink}href'
+        element.attrib[href] = re.sub(r'\.\w+$', '.jpg', element.attrib[href])
+        element.attrib['mime-subtype'] = 'jpeg'
+
+    # upload modified document
+    s3.put_object(
+        Bucket=DESTINATION_BUCKET,
+        Key=s3_key,
```
giorgiosironi: S3 in general has [eventual consistency on updates of keys](https://docs.aws.amazon.com/AmazonS3/latest/dev/Introduction.html#ConsistencyModel); so renaming would be fine (a new key is created), the overwrite may not (clients continuing to see the older content).
The way to treat keys to avoid these problems is to think of them as immutable, hence `s3-key` should be a new value like `elife-666-converted-images.xml` or any other suffix. This requires, however, to pass the new file name to the next task so that it does not assume what the file name is. Will be interesting to see how Airflow allows multiple tasks (the original extraction and this) to store a value for a downstream task.

dags/process_elife_zip_dag.py
GiancarloFusiello: `is_image` removed. Agreed to only convert tiff files.

config/airflow.dev.cfg
GiancarloFusiello: Should this still be a configurable option or should I just use `/tmp` all the time?

dags/process_elife_zip_dag.py
GiancarloFusiello: to quote the python documentation:
```If dir is not None, the file will be created in that directory; otherwise, a default directory is used. The default directory is chosen from a platform-dependent list, but the user of the application can control the directory location by setting the TMPDIR, TEMP or TMP environment variables.```

File: dags/process_elife_zip_dag.py (hunk 5 above)
Comment (GiancarloFusiello) [reply dags/process_elife_zip_dag.py]: I'll update to `temp_unzipped_file`

File: dags/process_elife_zip_dag.py (hunk 6 above)

File: dags/process_elife_zip_dag.py (hunk 8 above)
Comment (GiancarloFusiello) [reply dags/process_elife_zip_dag.py]: explanation added.

File: dags/process_elife_zip_dag.py [hunk 12]
```diff
@@ -138,30 +142,32 @@ def update_tiff_references_to_jpeg_in_article(**context):
...
+
+    return s3_key
 
 
 def wrap_article_in_libero_xml_and_send_to_service(**context):
-    zip_file_name = get_file_name_passed_to_dag_run_conf_file(**context)
-    article_name = get_expected_elife_article_name(zip_file_name)
-    s3_key = zip_file_name.replace('.zip', '/') + article_name
+    s3_key = get_return_value_from_previous_task(**context)
+    logger.info('ARTICLE S3 KEY PASSED FROM PREVIOUS TASK= %s', s3_key)
```
giorgiosironi: why the uppercase? Just standing out?

GiancarloFusiello: Correct. There are a lot of logs, especially if you enable debug mode.

PR #22, opened 2019-06-24
Remove related article tags

GiancarloFusiello:
In reference to https://github.com/libero/libero/issues/184

### DIFF
--- a/dags/process_elife_zip_dag.py
+++ b/dags/process_elife_zip_dag.py
@@ -57,6 +57,15 @@ def get_expected_elife_article_name(file_name: str) -> str:
     return article_name.group() + '.xml'
 
 
+def get_article_s3_key_from_previous_task(context: dict) -> str:
+    s3_key = get_return_value_from_previous_task(context)
+    logger.info('ARTICLE S3 KEY PASSED FROM PREVIOUS TASK= %s', s3_key)
+    previous_task = get_previous_task_name(context)
+    message = 'article s3 key was not passed from task %s' % previous_task
+    assert s3_key and isinstance(s3_key, str), message
+    return s3_key
+
+
 def extract_archived_files_to_bucket(**context):
     zip_file_name = get_file_name_passed_to_dag_run_conf_file(context)
     article_name = get_expected_elife_article_name(zip_file_name)
@@ -85,12 +94,7 @@ def extract_archived_files_to_bucket(**context):
                     Key=s3_key,
                     Body=temp_unzipped_file
                 )
-                logger.info(
-                    '%s uploaded to %s/%s',
-                    zipped_file_path,
-                    DESTINATION_BUCKET,
-                    folder_name
-                )
+                logger.info('%s uploaded to %s', s3_key, DESTINATION_BUCKET)
 
         # check if expected article in zip file
         if not [fn for fn in ZipFile(temp_zip_file).namelist() if article_name in fn]:
@@ -132,7 +136,7 @@ def convert_tiff_images_in_expanded_bucket_to_jpeg_images(**context):
                 logger.info('%s uploaded to %s',key, DESTINATION_BUCKET)
 
 
-def update_tiff_references_to_jpeg_in_article(**context):
+def update_tiff_references_to_jpeg_in_article(**context) -> str:
     zip_file_name = get_file_name_passed_to_dag_run_conf_file(context)
     article_name = get_expected_elife_article_name(zip_file_name)
     folder_name = zip_file_name.replace('.zip', '/')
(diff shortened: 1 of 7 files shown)


dags/process_elife_zip_dag.py
```diff
@@ -156,17 +160,38 @@ def update_tiff_references_to_jpeg_in_article(**context):
...
+    response = s3.get_object(Bucket=DESTINATION_BUCKET, Key=s3_key)
+    article_xml = etree.parse(BytesIO(response['Body'].read()))
+
+    matches = article_xml.xpath('//related-article')
+    if matches:
+        for element in matches:
+            element.getparent().remove(element)
+
+        # upload modified document
+        s3_key = s3_key.replace('.xml', '-remove_related_article.xml')
```
giorgiosironi: this is getting unwieldy, so we need to think a standard solution for names, what about an incremental number like `-edit_1`, `edit_2` so that it's clear in what order edits were performed?

GiancarloFusiello: I like the idea of a better naming convention but I don't think `edit_1`, `edit_2` is very descriptive.

Another solution is, instead of downloading and uploading edits to s3, the article bytes could be passed from one task to the next via the return statement? This would result in only the original article being in s3 and the final xml being sent to the content-store.

giorgiosironi: I think that solution would work well even with multiple workers (it's Airflow that passes things around for us) so interesting unless it creates tons of database rows as a result. We would reserve external storage like S3 for binary files whose size is on the order of MBs like images I guess.

dags/process_elife_zip_dag.py
```diff
@@ -57,8 +58,16 @@ def get_expected_elife_article_name(file_name: str) -> str:
     return article_name.group() + '.xml'
 
 
+def get_article_from_previous_task(context: dict) -> ElementTree:
+    article_bytes = get_return_value_from_previous_task(context)
+    previous_task = get_previous_task_name(context)
+    message = 'Article bytes were not passed from task %s' % previous_task
+    assert article_bytes and isinstance(article_bytes, bytes), message
```
giorgiosironi: wonder if we should introduce a stronger type, like a Value Object that wraps the bytes and can still be pickled?

tests/test_process_elife_zip_dag.py
```diff
@@ -45,60 +48,153 @@ def test_get_expected_elife_article_name_raises_exception_if_zip_name_is_malform
...
+
+@pytest.mark.parametrize('return_value', [b'', '', None])
+def test_get_article_from_previous_task_raises_exception(return_value, context):
+    # setup
+    add_return_value_from_previous_task(return_value, context)
+    message = 'Article bytes were not passed from task previous_task'
+    # setup
+    with pytest.raises(AssertionError) as error:
+        get_article_from_previous_task(context)
+        assert str(error.value) == message
```
giorgiosironi: here it's unclear whether `AssertionError` is raised by `get_article_from_previous_task` (which is what is being tested) or from the following `assert`?

giorgiosironi: Maybe it stems from the fact that the `assert` should be outside the `with`, otherwise it won't be reached?

tests/test_process_elife_zip_dag.py
```diff
@@ -45,60 +48,153 @@ def test_get_expected_elife_article_name_raises_exception_if_zip_name_is_malform
...
     keys = [folder_name + fn for fn in ZipFile(get_asset(file_name)).namelist()]
-    keys.append(folder_name)
+    keys = itertools.chain(keys, [folder_name])
     mocker.patch('dags.process_elife_zip_dag.list_bucket_keys_iter', return_value=keys)
 
+    # test
     convert_tiff_images_in_expanded_bucket_to_jpeg_images(**context)
+
+    # recreate expected uploaded file names
     zipped_files = [fn.replace('.tif', '.jpg')
```
giorgiosironi: a bit involved to have to calculate a complex expected value, what about relaxing these assertions and check just the number of files, plus a couple of examples from `s3_client.uploaded_files`?

tests/test_process_elife_zip_dag.py
GiancarloFusiello: fixed.

tests/test_process_elife_zip_dag.py
GiancarloFusiello: yeah sure. I'll update.

dags/process_elife_zip_dag.py
giorgiosironi: (more of a suggestion for later as this is getting quite big)

PR #24, opened 2019-06-26
fix xml base

GiancarloFusiello:
- improve tests
- improve libero settings

### DIFF
--- a/config/airflow.dev.cfg
+++ b/config/airflow.dev.cfg
@@ -2,10 +2,10 @@
 # url to override boto's default of aws s3. If blank will use boto default.
 boto_endpoint_url = http://localstack:4572
 
-# name of bucket where article archives are stored
+# name of bucket where article zip files are stored
 source_bucket_name = dev-jats-ingester-incoming
 
-# name of bucket to extract article archive files
+# name of bucket to extract article files
 destination_bucket_name = dev-jats-ingester-expanded
 
 # url of article assets i.e. external files referenced in an article
--- a/config/airflow.dev.cfg
+++ b/config/airflow.dev.cfg
@@ -1,13 +1,24 @@
 [libero]
-# s3
+# url to override boto's default of aws s3. If blank will use boto default.
 boto_endpoint_url = http://localstack:4572
-source_bucket = dev-jats-ingester-incoming
-destination_bucket = dev-jats-ingester-expanded
-base_url = https://%(destination_bucket)s.s3.amazonaws.com
+
+# name of bucket where article archives are stored
+source_bucket_name = dev-jats-ingester-incoming
+
+# name of bucket to extract article archive files
+destination_bucket_name = dev-jats-ingester-expanded
+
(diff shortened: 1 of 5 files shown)


config/airflow.dev.cfg
```diff
@@ -1,13 +1,24 @@
 [libero]
-# s3
+# url to override boto's default of aws s3. If blank will use boto default.
 boto_endpoint_url = http://localstack:4572
-source_bucket = dev-jats-ingester-incoming
-destination_bucket = dev-jats-ingester-expanded
-base_url = https://%(destination_bucket)s.s3.amazonaws.com
+
+# name of bucket where article archives are stored
```
giorgiosironi: `archive` is a bit of a bad word, just `artilcles are stored`?

config/airflow.dev.cfg
```diff
@@ -1,13 +1,24 @@
...
+source_bucket_name = dev-jats-ingester-incoming
+
+# name of bucket to extract article archive files
```
giorgiosironi: same here

config/airflow.dev.cfg
```diff
@@ -1,13 +1,24 @@
...
+destination_bucket_name = dev-jats-ingester-expanded
```
giorgiosironi: good suffixes, update `sample-configuration`

dags/process_elife_zip_dag.py
```diff
@@ -179,10 +179,10 @@ def wrap_article_in_libero_xml_and_send_to_service(**context) -> None:
 
     # add xml:base attribute to article element
     root = article_xml.getroot()
-    key = get_file_name_passed_to_dag_run_conf_file(context).replace('zip', '')
+    key = get_file_name_passed_to_dag_run_conf_file(context).replace('.zip', '')
```
giorgiosironi: this should remove `.zip` from the end, rather than from anywhere; or get the name in any other way

tests/test_process_elife_zip_dag.py
```diff
@@ -213,7 +213,11 @@ def test_wrap_article_in_libero_xml_and_send_to_service(context, s3_client, requ
     article = xml.xpath('//libero:item/jats:article', namespaces=namespaces)[0]
     assert article is not None
     assert len(article.getchildren()) > 0
-    assert article.attrib['{%s}base' % XML_NAMESPACE].endswith('/')
+
+    xml_base = article.attrib['{%s}base' % XML_NAMESPACE]
+    article_assets_url = configuration.conf.get('libero', 'article_assets_url')
+    expected = '%s/%s' % (article_assets_url, file_name.replace('.zip', '/'))
+    assert xml_base == expected
```
giorgiosironi: would hardcode `expected`, if it changes we update the test but it doesn't build expectations with the same code that it is testing

GiancarloFusiello:
referencing issue https://github.com/libero/libero/issues/184 as this bug fix is part of this issue.

PR #25, opened 2019-06-26
move put request to separate task

GiancarloFusiello:
In reference to https://github.com/libero/libero/issues/187

### DIFF
--- a/dags/process_elife_zip_dag.py
+++ b/dags/process_elife_zip_dag.py
@@ -165,7 +165,7 @@ def strip_related_article_tags_from_article_xml(**context) -> bytes:
     return etree.tostring(article_xml, xml_declaration=True, encoding='UTF-8')
 
 
-def wrap_article_in_libero_xml_and_send_to_service(**context) -> None:
+def wrap_article_in_libero_xml_and_send_to_service(**context) -> bytes:
     article_xml = get_article_from_previous_task(context)
 
     # get article id
@@ -197,14 +197,25 @@ def wrap_article_in_libero_xml_and_send_to_service(**context) -> None:
         root
     )
 
+    return etree.tostring(xml, xml_declaration=True, encoding='UTF-8')
+
+
+def send_article_to_content_service(**context) -> None:
+    libero_xml = get_article_from_previous_task(context)
+
+    # get article id
+    xpath = '//libero:item/jats:article/jats:front/jats:article-meta/jats:article-id[@pub-id-type="publisher-id"]'
+    namespaces = {'libero': 'http://libero.pub', 'jats': 'http://jats.nlm.nih.gov'}
+    article_id = libero_xml.xpath(xpath, namespaces=namespaces)[0].text
+
     # make PUT request to service
     response = requests.put(
         '%s/items/%s/versions/1' % (SERVICE_URL, article_id),
-        data=etree.tostring(xml, xml_declaration=True, encoding='UTF-8'),
+        data=etree.tostring(libero_xml, xml_declaration=True, encoding='UTF-8'),
         headers={'content-type': 'application/xml'}
     )
     response.raise_for_status()
-    logger.info('RESPONSE= %s: %s', response.status_code, response.text)
+    logger.info('RESPONSE STATUS= %s', response.status_code)
 
 
 # schedule_interval is None because DAG is only run when triggered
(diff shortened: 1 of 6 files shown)


dags/process_elife_zip_dag.py
```diff
@@ -197,14 +197,28 @@ def wrap_article_in_libero_xml_and_send_to_service(**context) -> None:
...
     )
 
+    return etree.tostring(xml, xml_declaration=True, encoding='UTF-8')
+
+
+def send_article_to_content_service(**context) -> None:
+    libero_xml = get_article_from_previous_task(context)
+
+    # get article id
+    xpath = '//libero:item/jats:article/jats:front/jats:article-meta/jats:article-id[@pub-id-type="publisher-id"]'
```
giorgiosironi: is this supremely complex xpath only used here? sounds like a constant to put at the top of the module for reuse?

dags/process_elife_zip_dag.py
```diff
@@ -197,14 +197,28 @@ def wrap_article_in_libero_xml_and_send_to_service(**context) -> None:
...
+    namespaces = {'libero': 'http://libero.pub', 'jats': 'http://jats.nlm.nih.gov'}
```
giorgiosironi: same for namespaces, constants?

dags/process_elife_zip_dag.py
```diff
@@ -197,14 +197,28 @@ def wrap_article_in_libero_xml_and_send_to_service(**context) -> None:
         root
     )
 
+    return etree.tostring(xml, xml_declaration=True, encoding='UTF-8')
```
giorgiosironi: I see this is not explicitly logged but I trust it would be accessible in someway in the Airflow UI as it's the return value of a task

dags/process_elife_zip_dag.py
```diff
@@ -197,14 +197,28 @@ def wrap_article_in_libero_xml_and_send_to_service(**context) -> None:
...
-        data=etree.tostring(xml, xml_declaration=True, encoding='UTF-8'),
+        data=etree.tostring(libero_xml, xml_declaration=True, encoding='UTF-8'),
         headers={'content-type': 'application/xml'}
     )
     response.raise_for_status()
-    logger.info('RESPONSE= %s: %s', response.status_code, response.text)
+    logger.info('Libero wrapped article %s sent to %s with status code %s',
+                article_id,
+                SERVICE_URL,
+                response.status_code)
```
giorgiosironi: would keep the `response.text`; it's probably not helpful now but it could be in general e.g. being a [problem detail](https://tools.ietf.org/html/rfc7807) response

dags/process_elife_zip_dag.py
```diff
@@ -165,7 +165,7 @@ def strip_related_article_tags_from_article_xml(**context) -> bytes:
     return etree.tostring(article_xml, xml_declaration=True, encoding='UTF-8')
 
 
-def wrap_article_in_libero_xml_and_send_to_service(**context) -> None:
+def wrap_article_in_libero_xml_and_send_to_service(**context) -> bytes:
```
giorgiosironi: this does not "send to the service" anymore, so rename?

dags/process_elife_zip_dag.py
GiancarloFusiello: Correct, only used here in this context.

dags/process_elife_zip_dag.py
GiancarloFusiello: Correct, task return items are logged also.

dags/process_elife_zip_dag.py
GiancarloFusiello: the text can be seen in the logs for the previous task. This task is simply taking that text and sending it to the content store.

dags/process_elife_zip_dag.py
GiancarloFusiello: Only used in this test.

dags/process_elife_zip_dag.py

dags/process_elife_zip_dag.py
giorgiosironi: I'm not sure what the API returns as a response body, so ok with not logging possible garbage for now

GiancarloFusiello: response body returns the data that was sent at the moment.

PR #26, opened 2019-06-26
Fix problematic tiff conversions

GiancarloFusiello:
In reference to https://github.com/libero/libero/issues/189

- replace pillow library with wand. test with multiple zips

### DIFF
--- a/dags/process_elife_zip_dag.py
+++ b/dags/process_elife_zip_dag.py
@@ -18,7 +18,6 @@ from airflow.utils import timezone
 from lxml import etree
 from lxml.builder import ElementMaker
 from lxml.etree import ElementTree
-# from PIL import Image
 from wand.image import Image
 
 from aws import get_aws_connection, list_bucket_keys_iter
--- a/dags/process_elife_zip_dag.py
+++ b/dags/process_elife_zip_dag.py
@@ -18,7 +18,8 @@ from airflow.utils import timezone
 from lxml import etree
 from lxml.builder import ElementMaker
 from lxml.etree import ElementTree
-from PIL import Image
+# from PIL import Image
+from wand.image import Image
 
 from aws import get_aws_connection, list_bucket_keys_iter
 from task_helpers import (
@@ -120,12 +121,14 @@ def convert_tiff_images_in_expanded_bucket_to_jpeg_images(**context) -> None:
                     Key=key,
                     Fileobj=temp_tiff_file
                 )
+                temp_tiff_file.seek(0)
 
-                # tiff images are typically RGBA
-                # PIL.JpegImagePlugin.RAWMODE contains modes that can be saved as jpeg
-                # In order to convert from tiff we have to remove the alpha channel
+                logger.info('converting %s to jpeg', key)
+
                 temp_jpeg = BytesIO()
-                Image.open(temp_tiff_file).convert(mode='RGB').save(temp_jpeg, format='JPEG')
+                with Image(blob=temp_tiff_file) as img:
+                    img.format = 'jpeg'
+                    img.save(file=temp_jpeg)
 
                 key = re.sub(r'\.\w+$', '.jpg', key)
                 s3.put_object(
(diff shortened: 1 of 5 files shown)


dags/process_elife_zip_dag.py
```diff
@@ -120,12 +121,14 @@ def convert_tiff_images_in_expanded_bucket_to_jpeg_images(**context) -> None:
...
+                logger.info('converting %s to jpeg', key)
 
-                # tiff images are typically RGBA
-                # PIL.JpegImagePlugin.RAWMODE contains modes that can be saved as jpeg
-                # In order to convert from tiff we have to remove the alpha channel
                 temp_jpeg = BytesIO()
-                Image.open(temp_tiff_file).convert(mode='RGB').save(temp_jpeg, format='JPEG')
+                with Image(blob=temp_tiff_file) as img:
+                    img.format = 'jpeg'
+                    img.save(file=temp_jpeg)
```
giorgiosironi: `img.destroy()` required?

GiancarloFusiello: not that I can see in the docs. Also I'm saving to a bytes stream which, being jpeg, should only be in the kilobytes.

tests/test_process_elife_zip_dag.py
```diff
@@ -91,20 +91,24 @@ def test_extract_archived_files_to_bucket_raises_exception_when_article_not_in_z
     assert str(error.value) == 'elife-00666.xml not in elife-00666-vor-r1.zip: []'
 
 
-def test_convert_tiff_images_in_expanded_bucket_to_jpeg_images_using_article_with_tiff_images(context, s3_client, mocker):
+@pytest.mark.parametrize('zip_file, uploaded_file, num_of_tiffs', [
+    ('elife-36842-vor-r3.zip', 'elife-36842-vor-r3/elife-36842-fig1.jpg', 25),
+    ('elife-40092-vor-r2.zip', 'elife-40092-vor-r2/elife-40092-fig1.jpg', 11)
+])
+def test_convert_tiff_images_in_expanded_bucket_to_jpeg_images_using_article_with_tiff_images(
+        zip_file, uploaded_file, num_of_tiffs, context, s3_client, mocker):
```
giorgiosironi: `sample_uploaded_file` since its just one of 25?

GiancarloFusiello: updated.

dags/process_elife_zip_dag.py
giorgiosironi: there's some memory required for reading the TIFF, but not sure how garbage collection will work here without experimenting

giorgiosironi: there's some memory required for reading the TIFF, but not sure how garbage collection will work here without experimentinghttps://sorry-wand.readthedocs.io/en/latest/wand/image.html#wand.image.Image.destroy seems to say calling `destroy` allows garbage collection to work _after_ a context manager

GiancarloFusiello: the tiff file is a temp file so the context manager should close and delete the file for each tiff after converting.

GiancarloFusiello: > wand.readthedocs.io/en/latest/wand/image.html#wand.image.Image.destroy

The documentation here says you can call it manually. Looking at the source code, I can see that destroy is called using a context manager.

In wand.image:
```python
class Image(BaseImage):
    def close(self):
        """Closes the image explicitly. If you use the image object in
        :keyword:`with` statement, it was called implicitly so don't have to
        call it.

        .. note::

           It has the same functionality of :attr:`destroy()` method.

        """
        self.destroy()
```

```python
class BaseImage(Resource):
    def destroy(self):
        """Manually remove :class:`~.sequence.SingleImage`'s in
        the :class:`~.sequence.Sequence`, allowing it to
        be properly garbage collected after using a ``with Image()`` context
        manager.

        """
        while self.sequence:
            self.sequence.pop()
        super(Image, self).destroy()
```

```python
class Resource(object):
    def __enter__(self):
        return self

    def __exit__(self, type, value, traceback):
        self.destroy()
```

PR #28, opened 2019-06-27
use multiprocessing to speed up image conversion

GiancarloFusiello:


dags/process_elife_zip_dag.py
```diff
@@ -103,39 +131,19 @@ def extract_archived_files_to_bucket(**context) -> str:
...
-
-                key = re.sub(r'\.\w+$', '.jpg', key)
-                s3.put_object(
-                    Bucket=DESTINATION_BUCKET,
-                    Key=key,
-                    Body=temp_jpeg.getvalue()
-                )
-                logger.info('%s uploaded to %s',key, DESTINATION_BUCKET)
+    with multiprocessing.Pool() as p:
+        return p.map(convert_image_in_s3_to_jpeg, tiffs, chunksize=3)
```
giorgiosironi: is `3` the size of the pool? It has implications for memory as the more concurrent processes are there the more memory needs to be allocated for each of them.

dags/process_elife_zip_dag.py
```diff
@@ -103,39 +131,19 @@ def extract_archived_files_to_bucket(**context) -> str:
...
-                    img.save(file=temp_jpeg)
-
-                key = re.sub(r'\.\w+$', '.jpg', key)
-                s3.put_object(
-                    Bucket=DESTINATION_BUCKET,
-                    Key=key,
-                    Body=temp_jpeg.getvalue()
-                )
-                logger.info('%s uploaded to %s',key, DESTINATION_BUCKET)
+    with multiprocessing.Pool() as p:
```
giorgiosironi: not sure if this is idiomatic for Airflow; if we have a set of parallel processes shouldn't we execute a set of parallel tasks in the same DAG and join them after they have all finished? Would also not have the slight overhead of process creation and termination.
But it could be that the number of parallel branches has to be predetermined at DAG construction, that would make it viable as the number of images is only known way into the DAG execution. More investigation in this area?

dags/process_elife_zip_dag.py
GiancarloFusiello: no. The pool size is equivalent to the number or cores. `chunksize` refers to the number of items to pass to a process before spawning a new process. This means that if `tiffs` has <= 3 items only one process will be spawned.

dags/process_elife_zip_dag.py
```diff
@@ -66,6 +68,32 @@ def get_article_from_previous_task(context: dict) -> ElementTree:
...
+        with Image(file=temp_tiff_file) as img:
+            img.format = 'jpeg'
+            img.save(file=temp_jpeg)
+
+        key = re.sub(r'\.\w+$', '.jpg', key)
+        s3.put_object(
+            Bucket=DESTINATION_BUCKET,
+            Key=key,
+            Body=temp_jpeg.getvalue()
+        )
```
giorgiosironi: would introduce a log at the start and at the end of the method, that can help have a rough understanding of start and stop time for each section; when they execute in parallel it can be a bit of a mess to identify which process is doing what.

dags/process_elife_zip_dag.py
giorgiosironi: Ok, "chunk" means "batch" size here

PR #29, opened 2019-07-01
send logs to s3 bucket

GiancarloFusiello:


Dockerfile
```diff
@@ -1,6 +1,7 @@
 FROM python:3.7.3-slim as base
 
 ENV AIRFLOW_HOME=/airflow
+ENV AIRFLOW_CONN_REMOTE_LOGS=""
```
giorgiosironi: should add a comment on the side-effect of creating a connection id

docker-compose.yml
```diff
@@ -58,6 +58,7 @@ services:
       - db
     environment:
       WAIT_HOSTS: db:5432
+      AIRFLOW_CONN_REMOTE_LOGS: http://localstack:4572?host=http://localstack:4572
```
giorgiosironi: awful repeated value, but for local usage so...

tests/conftest.py
```diff
@@ -64,4 +64,4 @@ def s3_client(mocker):
     """
     mocks boto client
     """
-    return mocker.patch('boto3.client', new_callable=mocks.s3ClientMock)
+    return mocker.patch('airflow.hooks.S3_hook.S3Hook.get_conn', new_callable=mocks.s3ClientMock)
```
giorgiosironi: shouldn't we mock `get_s3_client`?

giorgiosironi: if it works good, if it doesn't don't do it

tests/test_aws.py
```diff
@@ -1,19 +1,27 @@
...
+    assert conn._endpoint._endpoint_prefix == 's3'
+    assert conn._endpoint.host == "https://s3.eu-west-2.amazonaws.com"
+
+
+def test_get_s3_client_using_AIRFLOW_CONN_env_variable():
+    # format: AIRFLOW_CONN_ prefix followed by 'core', 'remote_log_conn_id' setting
+    # found in the airflow.cfg
+    # For more information:
+    # https://airflow.apache.org/howto/connection/index.html#creating-a-connection-with-environment-variables
+    os.environ['AIRFLOW_CONN_REMOTE_LOGS'] = 'test-host:1234?host=http://test-host:1234'
```
giorgiosironi: either remove (actually: restore) the variable in the tear down to avoid interaction with other tests, in whatever way pytest supports this

tests/conftest.py
GiancarloFusiello: This is more complicated than it would seem due to python importing.

PR #30, opened 2019-07-02
fix jpeg extensions

GiancarloFusiello:
In reference to https://github.com/libero/libero/issues/200

### DIFF
--- a/dags/process_elife_zip_dag.py
+++ b/dags/process_elife_zip_dag.py
@@ -147,15 +147,16 @@ def update_tiff_references_to_jpeg_in_article(**context) -> bytes:
     article_name = get_expected_elife_article_name(zip_file_name)
     prefix = zip_file_name.replace('.zip', '/')
     s3_key = prefix + article_name
+
     s3 = get_s3_client()
     response = s3.get_object(Bucket=DESTINATION_BUCKET, Key=s3_key)
+
     article_bytes = BytesIO(response['Body'].read())
     article_xml = etree.parse(article_bytes)
-    matches = article_xml.xpath('//*[@mimetype="image" and @mime-subtype="tiff"]')
-    if matches:
-        for element in matches:
-            element.attrib[XLINK_HREF] = re.sub(r'\.\w+$', '.jpg', element.attrib[XLINK_HREF])
-            element.attrib['mime-subtype'] = 'jpeg'
+
+    for element in article_xml.xpath('//*[@mimetype="image" and @mime-subtype="tiff"]'):
+        element.attrib[XLINK_HREF] = re.sub(r'\.\w+$', '.jpg', element.attrib[XLINK_HREF])
+        element.attrib['mime-subtype'] = 'jpeg'
 
     return etree.tostring(article_xml, xml_declaration=True, encoding='UTF-8')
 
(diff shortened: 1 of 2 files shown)


dags/process_elife_zip_dag.py
```diff
@@ -79,7 +83,7 @@ def extract_archived_files_to_bucket(**context) -> str:
         )
         logger.info('ZIPPED FILES= %s', ZipFile(temp_zip_file).namelist())
 
-        folder_name = zip_file_name.replace('.zip', '/')
+        prefix = zip_file_name.replace('.zip', '/')
```
giorgiosironi: ok, more generic variable name

tests/test_process_elife_zip_dag.py
```diff
@@ -154,6 +155,39 @@ def test_update_tiff_references_to_jpeg_in_articles_using_article_without_tiff_r
...
+    assert len(article_xml.xpath('//*[@xlink:href="elife-40092-resp-fig1.jpg"]', namespaces=namespaces )) == 0
+
+    # test
+    returned_xml = add_missing_jpeg_extensions_in_article(**context)
+    returned_xml = etree.parse(BytesIO(returned_xml))
+    assert len(returned_xml.xpath('//*[@xlink:href="elife-40092-resp-fig1"]', namespaces=namespaces)) == 0
+    assert len(returned_xml.xpath('//*[@xlink:href="elife-40092-resp-fig1.jpg"]', namespaces=namespaces)) == 1
+
+
+def test_add_missing_jpeg_extensions_in_article_without_missing_jpeg_extension(context, s3_client):
```
giorgiosironi: good test cases

dags/process_elife_zip_dag.py
```diff
@@ -141,22 +145,44 @@ def convert_tiff_images_in_expanded_bucket_to_jpeg_images(**context) -> None:
...
         for element in matches:
-            href = '{http://www.w3.org/1999/xlink}href'
-            element.attrib[href] = re.sub(r'\.\w+$', '.jpg', element.attrib[href])
+            element.attrib[XLINK_HREF] = re.sub(r'\.\w+$', '.jpg', element.attrib[XLINK_HREF])
             element.attrib['mime-subtype'] = 'jpeg'
 
     return etree.tostring(article_xml, xml_declaration=True, encoding='UTF-8')
 
 
+def add_missing_jpeg_extensions_in_article(**context) -> bytes:
```
giorgiosironi: I think "missing jpeg" describes correctly a link that we know to be a JPEG (because of other attributes) but doesn't have an extension

thewilkybarkid:
Really not sure we should be making code changes to fix broken input (I would expect these cases to fail).

dags/process_elife_zip_dag.py
```diff
@@ -141,22 +145,44 @@ def convert_tiff_images_in_expanded_bucket_to_jpeg_images(**context) -> None:
...
+    zip_file_name = get_file_name_passed_to_dag_run_conf_file(context)
+    article_name = get_expected_elife_article_name(zip_file_name)
+    prefix = zip_file_name.replace('.zip', '/')
+    s3_key = prefix + article_name
+    s3 = get_s3_client()
+    response = s3.get_object(Bucket=DESTINATION_BUCKET, Key=s3_key)
```
giorgiosironi: I thought we were getting the XML from the previous activity? Getting a bit confused on when we download it (should be only one place) and when we get it through Xcom.

dags/process_elife_zip_dag.py
```diff
@@ -141,22 +145,44 @@ def convert_tiff_images_in_expanded_bucket_to_jpeg_images(**context) -> None:
...
+    article_bytes = BytesIO(response['Body'].read())
+    article_xml = etree.parse(article_bytes)
+    matches = article_xml.xpath(
+        '//*[@mimetype="image" and @mime-subtype="jpeg"]')
+    if matches:
```
giorgiosironi: what's the negative value of `matches`? If it's something like `[]` only the `for` (on an empty list) is needed rather than also the `if`?

GiancarloFusiello:
> Really not sure we should be making code changes to fix broken input (I would expect these cases to fail).

@thewilkybarkid had a chat with @BlueReZZ and he suggested that it should be done here but I agree, I think it should intentionally fail (but that's not good for demoing content ;). Perhaps a larger discussion is needed.

dags/process_elife_zip_dag.py
```diff
@@ -141,22 +145,44 @@ def convert_tiff_images_in_expanded_bucket_to_jpeg_images(**context) -> None:
...
+        for element in matches:
+            image_file_name = element.attrib[XLINK_HREF]
+            has_extension = re.search(r'\.\w+$', image_file_name)
+            if has_extension and not image_file_name.endswith('.jpg'):
```
giorgiosironi: do we have cases where the extension is obviously wrong and needs to be substituted or was it only a case of not having it?

GiancarloFusiello: removed. not sure what I was thinking.

dags/process_elife_zip_dag.py
GiancarloFusiello: removed. not sure what I was thinking.

dags/process_elife_zip_dag.py
GiancarloFusiello: no you're right. I've just rushed it. fixing.

dags/process_elife_zip_dag.py
GiancarloFusiello: Added test case.

dags/process_elife_zip_dag.py
```diff
@@ -141,28 +145,39 @@ def convert_tiff_images_in_expanded_bucket_to_jpeg_images(**context) -> None:
...
+    return etree.tostring(article_xml, xml_declaration=True, encoding='UTF-8')
+
+
+def add_missing_jpeg_extensions_in_article(**context) -> bytes:
+    article_xml = get_article_from_previous_task(context)
+    for element in article_xml.xpath('//*[@mimetype="image" and @mime-subtype="jpeg"]'):
+        image_file_name = element.attrib[XLINK_HREF]
+        has_extension = re.search(r'\.\w+$', image_file_name)
+        if has_extension and not image_file_name.endswith('.jpg'):
+            element.attrib[XLINK_HREF] = re.sub(r'\.\w+$', '.jpg', image_file_name)
```
giorgiosironi: make sure this case is necessary, otherwise get rid of it?

GiancarloFusiello: removed

PR #32, opened 2019-07-03
improve logic for finding the jats article xml file name in a zip file

GiancarloFusiello:
In reference to https://github.com/libero/libero/issues/203

### DIFF
--- a/dags/process_elife_zip_dag.py
+++ b/dags/process_elife_zip_dag.py
@@ -7,7 +7,7 @@ import re
 from datetime import timedelta
 from io import BytesIO
 from pathlib import Path
-from tempfile import TemporaryFile
+from tempfile import TemporaryFile, NamedTemporaryFile
 from xml.dom import XML_NAMESPACE
 from zipfile import ZipFile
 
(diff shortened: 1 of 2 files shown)


dags/process_elife_zip_dag.py
```diff
@@ -53,14 +53,28 @@
...
+        s3 = get_s3_client()
+        s3.download_fileobj(
+            Bucket=SOURCE_BUCKET,
+            Key=zip_file_name,
+            Fileobj=temp_zip_file
+        )
+
+        for zipped_file in ZipFile(temp_zip_file).namelist():
+            if zipped_file.endswith('.xml'):
+                xml = etree.parse(BytesIO(ZipFile(temp_zip_file).read(zipped_file)))
```
giorgiosironi: should create `ZipFile(temp_zip_file)` in a single place?

dags/process_elife_zip_dag.py
```diff
@@ -145,7 +149,7 @@ def convert_tiff_images_in_expanded_bucket_to_jpeg_images(**context) -> None:
 
 def update_tiff_references_to_jpeg_in_article(**context) -> bytes:
     zip_file_name = get_file_name_passed_to_dag_run_conf_file(context)
-    article_name = get_expected_elife_article_name(zip_file_name)
+    article_name = get_article_file_name(zip_file_name)
```
giorgiosironi: you wanted to improve the interface of this function to return the XML content directly?

dags/process_elife_zip_dag.py

dags/process_elife_zip_dag.py
GiancarloFusiello: correct. updated.

PR #33, opened 2019-07-03
Implement concurrent processing for expensive operations

GiancarloFusiello:
In reference to https://github.com/libero/libero/issues/204

### DIFF
--- a/dags/process_elife_zip_dag.py
+++ b/dags/process_elife_zip_dag.py
@@ -158,7 +158,7 @@ def extract_archived_files_to_bucket(**context) -> List[str]:
         # extract zip to cloud bucket
         thread_options = {
             'max_workers': WORKERS,
-            'thread_name_prefix': 'extracting_%s' % zip_file_name
+            'thread_name_prefix': 'extracting_%s_' % zip_file_name
         }
         with ThreadPoolExecutor(**thread_options) as p:
             prefix = zip_file_name.replace('.zip', '/')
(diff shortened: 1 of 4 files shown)


dags/process_elife_zip_dag.py
```diff
@@ -4,10 +4,12 @@
 """
 import logging
 import re
+from concurrent.futures.thread import ThreadPoolExecutor
```
giorgiosironi: I've used this to [parallelize loading HTTP images and similar in end2end testing](https://github.com/elifesciences/elife-spectrum/blob/54626dae48673056febdced18ca9505f44c5a30a/spectrum/checks.py#L907)

dags/process_elife_zip_dag.py
```diff
@@ -86,62 +146,31 @@ def extract_archived_files_to_bucket(**context) -> str:
...
-                temp_unzipped_file.seek(0)
-
-                s3_key = prefix + zipped_file_path
-                s3.put_object(
-                    Bucket=DESTINATION_BUCKET,
-                    Key=s3_key,
-                    Body=temp_unzipped_file
-                )
-                logger.info('%s uploaded to %s', s3_key, DESTINATION_BUCKET)
+        with ThreadPoolExecutor() as p:
```
giorgiosironi: need to be able to configure the `max_workers` argument, to tweak the number of threads. The [default](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor) seems to be `processors*5` which can be a lot.

dags/process_elife_zip_dag.py
```diff
@@ -86,62 +146,31 @@ def extract_archived_files_to_bucket(**context) -> str:
...
-
-                key = re.sub(r'\.\w+$', '.jpg', key)
-                s3.put_object(
-                    Bucket=DESTINATION_BUCKET,
-                    Key=key,
-                    Body=temp_jpeg.getvalue()
-                )
-                logger.info('%s uploaded to %s',key, DESTINATION_BUCKET)
-
+    with ThreadPoolExecutor() as p:
```
giorgiosironi: same here, can be a global configuration rather than being specific for the task

dags/process_elife_zip_dag.py
giorgiosironi: consider specifying `thread_name_prefix` which makes debugging easier; the Python threads are mapped on OS threads so they should be visible in `htop` with the right options

dags/process_elife_zip_dag.py

PR #34, opened 2019-07-03
add logic to not trigger dag for file already being processed

GiancarloFusiello:
In reference to https://github.com/libero/libero/issues/207

### DIFF
--- a/dags/trigger_dag.py
+++ b/dags/trigger_dag.py
@@ -9,11 +9,14 @@ from uuid import uuid4
 
 from airflow import DAG, configuration
 from airflow.api.common.experimental.trigger_dag import trigger_dag
+from airflow.models import DagRun
 from airflow.operators import python_operator
+from airflow.settings import Session
 from airflow.utils import timezone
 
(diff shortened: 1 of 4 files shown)


dags/trigger_dag.py
```diff
@@ -63,13 +66,24 @@ def run_dag_for_each_file(dag_to_trigger, **context) -> None:
     message = 'None type passed from previous task. Accepted types are set, list or tuple.'
     assert file_names is not None, message
 
+    session = Session()
+    files_triggered = []
     for file_name in file_names:
+
+        # check if a file has already been triggered for processing
+        if session.query(DagRun).filter(and_(DagRun.run_id.contains(file_name),
```
giorgiosironi: is `contains` strict enough?

giorgiosironi: you could have clashes like `23.zip` and `123.zip`?

giorgiosironi: `contains("_{file_name}")`?

PR #35, opened 2019-07-04
Improve error logs

GiancarloFusiello:
In reference to https://github.com/libero/libero/issues/208

- add response text to log when sending the article to the content store fails

### DIFF
--- a/dags/process_elife_zip_dag.py
+++ b/dags/process_elife_zip_dag.py
@@ -20,6 +20,7 @@ from airflow.utils import timezone
 from lxml import etree
 from lxml.builder import ElementMaker
 from lxml.etree import ElementTree
+from requests import HTTPError
 from wand.image import Image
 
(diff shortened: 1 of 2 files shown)


dags/process_elife_zip_dag.py
```diff
@@ -261,7 +262,13 @@ def send_article_to_content_service(**context) -> None:
         data=etree.tostring(libero_xml, xml_declaration=True, encoding='UTF-8'),
         headers={'content-type': 'application/xml'}
     )
-    response.raise_for_status()
+
+    try:
+        response.raise_for_status()
+    except HTTPError as e:
+        logger.error('%s HTTP Error= %s', response.status_code, response.text)
```
giorgiosironi: could include something about the request here (e..g the URL) to help correlate

GiancarloFusiello: yep that information will be logged when re-raising the error.

PR #36, opened 2019-07-04
fix missing protocols

GiancarloFusiello:
In reference to https://github.com/libero/publisher/issues/209

### DIFF
--- a/dags/process_elife_zip_dag.py
+++ b/dags/process_elife_zip_dag.py
@@ -335,7 +335,7 @@ strip_related_article_tags = python_operator.PythonOperator(
     dag=dag
 )
 
-add_missing_protocols_task = python_operator.PythonOperator(
+add_missing_uri_schemes_task = python_operator.PythonOperator(
     task_id='add_missing_uri_schemes',
     provide_context=True,
     python_callable=add_missing_uri_schemes,
@@ -367,7 +367,7 @@ extract_zip_files.set_downstream(convert_tiff_images)
 convert_tiff_images.set_downstream(update_tiff_references)
 update_tiff_references.set_downstream(add_missing_jpeg_extensions)
 add_missing_jpeg_extensions.set_downstream(strip_related_article_tags)
-strip_related_article_tags.set_downstream(add_missing_protocols_task)
-add_missing_protocols_task.set_downstream(wrap_article)
+strip_related_article_tags.set_downstream(add_missing_uri_schemes_task)
+add_missing_uri_schemes_task.set_downstream(wrap_article)
 wrap_article.set_downstream(send_article)
 send_article.set_downstream(reindex_search)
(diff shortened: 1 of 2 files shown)


dags/process_elife_zip_dag.py
```diff
@@ -212,6 +220,15 @@ def strip_related_article_tags_from_article_xml(**context) -> bytes:
     return etree.tostring(article_xml, xml_declaration=True, encoding='UTF-8')
 
 
+def add_missing_protocols(**context) -> bytes:
```
giorgiosironi: would keep the URI terminology and call these `schemes` rather than `protocols`: https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Examples

GiancarloFusiello: updated

dags/process_elife_zip_dag.py
```diff
@@ -315,6 +335,13 @@ def send_post_request_to_reindex_search_service():
     dag=dag
 )
 
+add_missing_protocols_task = python_operator.PythonOperator(
```
giorgiosironi: similar renaming

tests/test_process_elife_zip_dag.py
```diff
@@ -205,10 +198,26 @@ def test_strip_related_article_tags_from_article_xml_using_article_without_relat
...
-    return_value = strip_related_article_tags_from_article_xml(**context)
+    return_value = pezd.strip_related_article_tags_from_article_xml(**context)
     assert return_value == article_xml
 
 
+def test_add_missing_uri_schemes(context):
+    # setup
+    test_asset_path = str(get_asset('elife-36842.xml').absolute())
+    article_xml = etree.parse(test_asset_path)
+    xpath = '//*[starts-with(@xlink:href, "www.")]'
```
giorgiosironi: `www.` seems very specific, but if it works for our examples it's good enough

dags/process_elife_zip_dag.py
```diff
@@ -212,6 +220,15 @@ def strip_related_article_tags_from_article_xml(**context) -> bytes:
     return etree.tostring(article_xml, xml_declaration=True, encoding='UTF-8')
 
 
+def add_missing_uri_schemes(**context) -> bytes:
+    article_xml = get_article_from_previous_task(context)
+    elements = article_xml.xpath('//*[starts-with(@xlink:href, "www.")]', namespaces=XLINK)
+    for element in elements:
+        element.attrib[XLINK_HREF] = 'http://' + element.attrib[XLINK_HREF]
```
giorgiosironi: `http` is probably the safest default wrt `https`, it is usually redirected to `https` if available

tests/test_process_elife_zip_dag.py
GiancarloFusiello: follow the bot as an example: https://github.com/elifesciences/bot-lax-adaptor/blob/58318286ab8d07d206abe34b313f63ce51117d31/src/main.py#L351-L370

PR #37, opened 2019-07-04
use minio in dev environment

GiancarloFusiello:
In reference to https://github.com/libero/publisher/issues/210

### DIFF
--- a/Dockerfile.awscli
+++ b/Dockerfile.awscli
@@ -1,4 +1,4 @@
-FROM imione/awscli-local
+FROM mikesir87/aws-cli:latest
 
 ADD https://github.com/ufoscout/docker-compose-wait/releases/download/2.5.0/wait /wait
 RUN chmod +x /wait
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,7 +7,6 @@ services:
     environment:
       MINIO_ACCESS_KEY: longkey
       MINIO_SECRET_KEY: verysecretkey
-      MINIO_REGION: eu-west-2
     ports:
       - 9000:9000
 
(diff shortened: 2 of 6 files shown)


Dockerfile.awscli
```diff
@@ -1,4 +1,4 @@
-FROM imione/awscli-local
+FROM mikesir87/aws-cli:latest
```
giorgiosironi: unrelated, right?

docker-compose.yml
```diff
@@ -1,16 +1,15 @@
...
+    command: server /data
     environment:
-      - SERVICES=s3
-      - PORT_WEB_UI=8081
-      - DEFAULT_REGION=eu-west-2
-      - AWS_ACCESS_KEY_ID=foo
-      - AWS_SECRET_ACCESS_KEY=bar
+      MINIO_ACCESS_KEY: longkey
+      MINIO_SECRET_KEY: verysecretkey
+      MINIO_REGION: eu-west-2
```
giorgiosironi: maybe mandatory to specify, but why `eu-west-2` if it's fake? would keep the default of defaults for AWS SDKs, `us-east-1`, if a value is needed here

docker-compose.yml
```diff
@@ -1,16 +1,15 @@
...
     ports:
-      - "8081:8081"
+      - 9000:9000
```
giorgiosironi: this is a YAML string, so whenever there is any non-alphanumeric character keep the `"` quotes around it for safety:
https://twitter.com/sistasonDE/status/1045353434394841088

docker-compose.yml
```diff
@@ -22,10 +21,12 @@ services:
       - ./tests/assets/:/src/app/assets/
     command: ./create-s3-buckets.sh
     depends_on:
-      - localstack
+      - s3
     environment:
-      WAIT_HOSTS: localstack:4572, localstack:4563
-      LOCALSTACK_HOST: localstack
+      WAIT_HOSTS: s3:9000
```
giorgiosironi: good simplification

Dockerfile.awscli
GiancarloFusiello: awscli-local is the localstack wrapper for aws cli. I'm now just using aws-cli.

PR #38, opened 2019-07-05
remove object id tags

GiancarloFusiello:
In reference to https://github.com/libero/publisher/issues/209

### DIFF
--- a/dags/process_elife_zip_dag.py
+++ b/dags/process_elife_zip_dag.py
@@ -221,6 +221,14 @@ def strip_related_article_tags_from_article_xml(**context) -> bytes:
     return etree.tostring(article_xml, xml_declaration=True, encoding='UTF-8')
 
 
+def strip_object_id_tags_from_article_xml(**context) -> bytes:
+    article_xml = get_article_from_previous_task(context)
+    for element in article_xml.xpath('//object-id'):
+        element.getparent().remove(element)
+
+    return etree.tostring(article_xml, xml_declaration=True, encoding='UTF-8')
+
+
 def add_missing_uri_schemes(**context) -> bytes:
     article_xml = get_article_from_previous_task(context)
     elements = article_xml.xpath('//*[starts-with(@xlink:href, "www.")]', namespaces=XLINK)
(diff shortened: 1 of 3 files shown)


dags/process_elife_zip_dag.py
```diff
@@ -221,6 +221,14 @@ def strip_related_article_tags_from_article_xml(**context) -> bytes:
     return etree.tostring(article_xml, xml_declaration=True, encoding='UTF-8')
 
 
+def strip_object_id_tags_from_article_xml(**context) -> bytes:
```
giorgiosironi: the implementation of this could be extracted into a private function called also from `strip_related_article_tags_from_article_xml`, since it's identical but for the xpath query

tests/test_process_elife_zip_dag.py
```diff
@@ -202,6 +202,35 @@ def test_strip_related_article_tags_from_article_xml_using_article_without_relat
     assert return_value == article_xml
 
 
+def test_strip_object_id_tags_from_article_xml_using_article_with_related_article_tag(context):
+    # setup
+    test_asset_path = str(get_asset('elife-36842.xml').absolute())
+    article_xml = etree.parse(test_asset_path)
+    xpath = '//object-id'
+    assert len(article_xml.xpath(xpath)) == 49
```
giorgiosironi: a bit of a magic number, could just be asserted to be greater than 0 and the test would still cover this case?

tests/test_process_elife_zip_dag.py
```diff
@@ -202,6 +202,35 @@ def test_strip_related_article_tags_from_article_xml_using_article_without_relat
...
+        return_value=etree.tostring(article_xml, xml_declaration=True, encoding='UTF-8'),
+        context=context
+    )
+    # test
+    return_value = pezd.strip_object_id_tags_from_article_xml(**context)
+    xml = etree.parse(BytesIO(return_value))
+    assert len(xml.xpath(xpath)) == 0
+
+
+def test_strip_object_id_tags_from_article_xml_using_article_without_related_article_tag(context):
```
giorgiosironi: ~why the two different test cases for this new activity, with and without related articles?~ actually reading the test, I think it's just the test names that need to be updated to say `(with|without)_object_id_tags`

tests/test_process_elife_zip_dag.py
GiancarloFusiello: updated

tests/test_process_elife_zip_dag.py
GiancarloFusiello: updated

dags/process_elife_zip_dag.py
GiancarloFusiello: turns out lxml provides this functionality. Updated.

PR #41, opened 2019-07-05
fix branching

GiancarloFusiello:
Before sending an article to the content store, the article is retrieved from the previous task. Since introducing branching, the task to retrieve the article from is now indeterministic. This PR adds a method to specify which upstream task to retrieve the article from.

### DIFF
--- a/tests/test_task_helpers.py
+++ b/tests/test_task_helpers.py
@@ -6,7 +6,7 @@ from dags.task_helpers import (
     get_file_name_passed_to_dag_run_conf_file
 )
 from tests.factories import TaskInstanceFactory
-from tests.helpers import add_return_value_from_previous_task
+from tests.helpers import populate_task_return_value
 
 
 def test_get_previous_task_name(context):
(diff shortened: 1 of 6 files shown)


tests/test_task_helpers.py
```diff
@@ -21,11 +22,21 @@ def test_get_previous_task_name_without_previous_task():
...
     expected = 'previous_return_value'
-    context['task_instance'].xcom_pull = lambda **kwargs: expected
+    add_return_value_from_previous_task(expected, context)
     result = get_return_value_from_previous_task(context)
     assert result == expected
 
 
+def test_get_return_value_from_branched_previous_task(branched_context):
+    expected = 'previous_return_value'
+    add_return_value_from_previous_task(expected, branched_context, task_id='branched_previous_task')
```
giorgiosironi: should it be named `returning_previous_task` or similar, rather than `branched_previous_task`? Maybe this name is abstract and reusable across tests...

giorgiosironi: would still keep the two branched task names similar, since there's no preferred one

PR #42, opened 2019-07-09
get correct task in test branch context fixture

GiancarloFusiello:
- simple update that makes the fixture more deterministic.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -88,7 +88,15 @@ def branched_context():
             session=session
         )
 
-    ti = dag_run.get_task_instances()[1]
+    ti = None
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -88,7 +88,15 @@ def branched_context():
...
         )
 
-    ti = dag_run.get_task_instances()[1]
+    ti = None
+    for instance in dag_run.get_task_instances():
+        if instance.task_id == 'current_task':
+            ti = instance
+            break
+
+    if ti is None:
```
giorgiosironi: also if you found more than one, right? (the `break` will stop at the first)

PR #44, opened 2019-07-11
parse response xml error message from content store to log message

GiancarloFusiello:
In reference to https://github.com/libero/publisher/issues/208

### DIFF
--- a/dags/process_elife_zip_dag.py
+++ b/dags/process_elife_zip_dag.py
@@ -294,11 +294,17 @@ def send_article_to_content_service(upstream_task_id: str = None, **context) -> 
 
     try:
         response.raise_for_status()
-    except HTTPError:
-        response_xml = etree.parse(BytesIO(response.content))
-        details = response_xml.find('{urn:ietf:rfc:7807}details')
-        logger.error('HTTPError: %s - %s', response.status_code, details.text)
-        raise
+    except HTTPError as http_error:
+        # parse the xml response for readability
+        try:
+            response_xml = etree.parse(BytesIO(response.content))
+            details = response_xml.find('{urn:ietf:rfc:7807}details')
+            details = getattr(details, 'text', '')
+        except etree.XMLSyntaxError:
+            details = response.text
+
+        logger.error('HTTPError: %s - %s', response.status_code, details)
+        raise http_error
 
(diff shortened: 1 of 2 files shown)


dags/process_elife_zip_dag.py
```diff
@@ -294,9 +294,11 @@ def send_article_to_content_service(upstream_task_id: str = None, **context) ->
 
     try:
         response.raise_for_status()
-    except HTTPError as e:
-        logger.error('%s HTTP Error= %s', response.status_code, response.text)
-        raise e
+    except HTTPError:
+        response_xml = etree.parse(BytesIO(response.content))
```
giorgiosironi: what happens if the XML is not parseable? In an imperfect world happens more often than we think, especially with gateway in the middle :sweat_smile: would still try to log the raw `response.content` string in that case

dags/process_elife_zip_dag.py
```diff
@@ -294,9 +294,11 @@ def send_article_to_content_service(upstream_task_id: str = None, **context) ->
...
+        details = response_xml.find('{urn:ietf:rfc:7807}details')
+        logger.error('HTTPError: %s - %s', response.status_code, details.text)
```
giorgiosironi: I agree with the logging level, any other field from the XML that would be useful to log? According to the RFC there is `title` and `type`

giorgiosironi: I saw the test example, `title` only then

dags/process_elife_zip_dag.py
GiancarloFusiello: If the response is not parsable then that would indicate that the schema for errors is not consistent and I would expect the system to fall over.

dags/process_elife_zip_dag.py
GiancarloFusiello: have a look at the test code for an example of the response structure. Also remember that I still raise the original exception which will provide details about the response.

dags/process_elife_zip_dag.py
giorgiosironi: I agree it should fall over, just wanted a way to know what the response was when it fell over, since it can be difficult to find out what happened.

PR #45, opened 2019-07-12
add support for biorxiv

GiancarloFusiello:
In reference to https://github.com/libero/publisher/issues/219

### DIFF
--- a/dags/process_zip_dag.py
+++ b/dags/process_zip_dag.py
@@ -92,6 +92,21 @@ def get_article_from_previous_task(context: dict, task_id: str = None) -> Elemen
     return etree.parse(BytesIO(article_bytes))
 
 
+def get_element_text_from_xpaths(xml: ElementTree, xpaths: List[str], namespaces: dict = None) -> str:
+    """
+    Searches an lxml ElementTree object for an element using one or more xpaths
+    and returns the text of the first element found.
+    """
+    text = None
+    for xpath in xpaths:
+        elements = xml.xpath(xpath, namespaces=namespaces)
+        if elements:
+            text = elements[0].text
+            break
+    assert text is not None, 'Xpaths not found in xml: %s' % xpaths
+    return text
+
+
 def extract_file_to_s3(args: tuple) -> str:
     """
     Reads a file contained in a zip file and uploads it to a predefined bucket.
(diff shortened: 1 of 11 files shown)


dags/process_zip_dag.py
```diff
@@ -201,6 +216,9 @@ def update_tiff_references_to_jpeg_in_article(**context) -> bytes:
         element.attrib[XLINK_HREF] = re.sub(r'\.\w+$', '.jpg', element.attrib[XLINK_HREF])
         element.attrib['mime-subtype'] = 'jpeg'
 
+    for element in article_xml.xpath('//*[contains(@xlink:href, ".tif")]', namespaces=XLINK):
+        element.attrib[XLINK_HREF] = re.sub(r'\.\w+$', '.jpg', element.attrib[XLINK_HREF])
```
giorgiosironi: consider extracting the two `for` loops into two functions to clarify what they are doing

tests/assets/__init__.py
```diff
@@ -1,9 +1,17 @@
...
-        raise FileNotFoundError
+        raise FileNotFoundError(name)
     return path
+
+
+def find_asset(name: str) -> List[Path]:
+    paths = list(Path('.').rglob(name))
+    if not paths:
+        raise FileNotFoundError(name)
+    return paths
```
giorgiosironi: about the `.meca` files, they need to be in Git LFS:
https://github.com/libero/jats-ingester/blob/master/.gitattributes

giorgiosironi: amend the original commit because removing a file and putting it in LFS makes `git diff` and other commands go crazy

tests/test_process_zip_dag.py
```diff
@@ -271,7 +293,7 @@ def test_wrap_article_in_libero_xml(context):
     assert len(article.getchildren()) > 0
 
     xml_base = article.attrib['{%s}base' % XML_NAMESPACE]
-    expected = 'https://test-expanded-bucket.test.com/elife-36842-vor-r3/'
+    expected = 'https://test-expanded-bucket.test.com/' + expected_suffix
```
giorgiosironi: would rename `expected_suffix` to `expected_path`, URL terminology rather than S3 terminology

tests/assets/__init__.py
GiancarloFusiello: updated.

tests/test_process_zip_dag.py
GiancarloFusiello: updated.

PR #47, opened 2019-08-08
add get article id strategies

GiancarloFusiello:
In reference to: https://github.com/libero/publisher/issues/231


dags/libero/xml/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+XLINK_MAP = {'xlink': XLINK_NS}
+
+# clark notation
+# reference: http://www.jclark.com/xml/xmlns.htm
+XLINK_HREF = '{%s}href' % XLINK_NS
+XML_BASE = '{%s}base' % XML_NAMESPACE
+
+
+
+def get_element_text_from_xpaths(xml: ElementTree, xpaths: Iterable[str], namespaces: dict = None) -> str:
```
giorgiosironi: annoying to have to pass in the `namespaces`, but necessary

dags/libero/xml/jats/xpaths.py
```diff
@@ -0,0 +1,12 @@
...
+RELATED_ARTICLE = '//related-article'
+OBJECT_ID = '//object-id'
+
+ARTICLE_ID_BY_PUBLISHER_ID = '/article/front/article-meta/article-id[@pub-id-type="publisher-id"]'
+ARTICLE_ID_NOT_BY_PMID_PMC_DOI = '/article/front/article-meta/article-id[not(@pub-id-type="pmid") and not(@pub-id-type="pmc") and not(@pub-id-type="doi")]'
+ARTICLE_ID_BY_ELOCATION_ID = '/article/front/article-meta/elocation-id'
+ARTICLE_ID_BY_DOI = '/article/front/article-meta/article-id[@pub-id-type="doi"]'
+
+IMAGE_BY_TIFF_MIMETYPE = '//*[@mimetype="image" and @mime-subtype="tiff"]'
+IMAGE_BY_JPEG_MIMETYPE = '//*[@mimetype="image" and @mime-subtype="jpeg"]'
```
giorgiosironi: I like that they are all kept together in a single place

dags/libero/xml/__init__.py
GiancarloFusiello: agreed but that's part of lxml. There is a default value of None in the case where a namespace is not required for a query.

PR #48, opened 2019-09-13
add clairvoyant airflow maintenance dags to docker image

GiancarloFusiello:
In reference to https://github.com/libero/publisher/issues/247

Downloads relevant [maintenance DAGs by teamclairvoyant](https://github.com/teamclairvoyant/airflow-maintenance-dags) and adds them to the `dags` directory under the directory `maintenance-dags`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,11 +30,9 @@ COPY ./dags ${AIRFLOW_HOME}/dags
 
 # add maintenance dags
 ADD https://raw.githubusercontent.com/teamclairvoyant/airflow-maintenance-dags/master/clear-missing-dags/airflow-clear-missing-dags.py \
-    ${AIRFLOW_HOME}/dags/maintenance-dags/airflow-clear-missing-dags.py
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -27,6 +27,15 @@ RUN pip install -U pip \
     && rm -rf /tmp/*
 
 COPY ./dags ${AIRFLOW_HOME}/dags
+
+# add maintenance dags
+ADD https://raw.githubusercontent.com/teamclairvoyant/airflow-maintenance-dags/master/clear-missing-dags/airflow-clear-missing-dags.py \
```
diversemix: Can we do this in one layer using:
```
ADD [--chown=<user>:<group>] ["<src>",... "<dest>"]
```
ref: https://docs.docker.com/engine/reference/builder/#add

GiancarloFusiello: updated.

diversemix:
:shipit:

PR #52, opened 2019-09-26
Get categories from jats xml

GiancarloFusiello:
In reference to https://github.com/libero/publisher/issues/245

With test cases implemented as per https://github.com/libero/publisher/issues/256

- Adds `get_categories` function to the `libero.xml.jats` library.


dags/libero/xml/libero/__init__.py
```diff
@@ -16,4 +16,4 @@ def get_content_id(jats_xml: ElementTree) -> str:
         xml=jats_xml,
         xpaths=[xpaths.ID],
         namespaces=LIBERO_MAP
-    )
+    )[0]
```
giorgiosironi: could assert this list has 1 element before returning the 0-th one?

tests/test_xml_jats.py
```diff
@@ -53,3 +53,136 @@ def test_get_article_id_raises_exception_when_id_not_found():
...
+     '          <subject>Housing</subject>'
+     '        </subj-group>'
+     '        <subj-group>'
+     '          <subject>General Economics</subject>'
+     '        </subj-group>'
+     '      </article-categories>'
+     '    </article-meta>'
+     '  </front>'
+     '</article>',
+     ['Cancer Biology', 'Data', 'Housing', 'General Economics']),
```
giorgiosironi: @FAtherden-eLife can you validate these tests work as intended?
We would also need a test with nested tags (such as italic?) within the subject name, which you mentioned before.

FAtherden-eLife: > @FAtherden-eLife can you validate these tests work as intended?

Yup, they look good.

> We would also need a test with nested tags (such as italic?) within the subject name, which you mentioned before.

Yes, good point. I neglected to include that. @GiancarloFusiello, another (more complex) test case is included [here](https://github.com/libero/publisher/issues/256#issuecomment-535876763).

FAtherden-eLife: @giorgiosironi, @GiancarloFusiello Do we also need to add a test case which covers the logic set out in the Notes section [here](https://docs.google.com/document/d/1PmLD6NEiqpjKTnjW-n1Hho_G-znp5lW3ba8PGmOHnQE/edit?ts=5d8b6ec9#heading=h.9911g2g4u32r)? i.e. two articles with the same content in `subject` except the case is different (e.g. `Ecology` and `ecology`), the first value being used and second taking the value from the first? Or will this be dealt with elsewhere?

GiancarloFusiello: @FAtherden-eLife I don't think we need a test case for the Notes scenario. Currently implementing the test case add [here](https://github.com/libero/publisher/issues/256) that deals with styling tags.

giorgiosironi: Agreed on elsewhere, that will be covered by the tests of the requests towards `scholarly-content-detail`; separate from extracting the values which is what we do here.

dags/libero/xml/libero/__init__.py
giorgiosironi: If I read that correctly, that asserts that the number of elements is `> 0` whereas here we should check that it is `== 1` to avoid silently discarding elements after the first

GiancarloFusiello: It's likely that there will be multiple results as multiple XPaths (strategies) can be used to derive the target element. As these are ordered, the first match is used as it should be higher in the order of preferred locations.

giorgiosironi: Ah it's intended then, :+1:

tests/test_xml_jats.py
giorgiosironi: @FAtherden-eLife I'll leave it to you to provide feedback on the extents of the tests that should be covered here, no further concerns from me

PR #53, opened 2019-10-01
replace requests-mock with responses library

GiancarloFusiello:
I encountered an issue serialising and deserialising python objects to json using `requests-mock`. This library is maintained by the Sentry team and **seems** a little more robust.

- replace `requests-mock` with `responses` library for mocking request responses

### DIFF
--- a/tests/test_process_zip_dag.py
+++ b/tests/test_process_zip_dag.py
@@ -4,6 +4,7 @@ from io import BytesIO
 from zipfile import ZipFile
 
 import pytest
+import responses
 from lxml import etree
 from requests.exceptions import HTTPError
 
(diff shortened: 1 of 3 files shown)


tests/test_process_zip_dag.py
```diff
@@ -311,38 +312,40 @@ def test_wrap_article_in_libero_xml_raises_exception_if_xml_path_not_returned_by
...
-              '<problem xmlns="urn:ietf:rfc:7807" xml:lang="en">'
-              '  <status>400</status>'
-              '  <title>Failed to load asset</title>'
-              '  <details>Failed to load https://unstable-jats-ingester-expanded.'
-              '  s3.amazonaws.com/elife-00666-vor-r1/10.7554/eLife.00666.004 '
-              '  due to "404 Not Found".'
-              '  </details>'
-              '</problem>')
+        status=400,
+        body=(b'<?xml version="1.0" encoding="UTF-8"?>'
```
giorgiosironi: a bit strange to have to specify bytes over a string?

GiancarloFusiello: good point. I did this following an example but the [documentation](https://github.com/getsentry/responses) does specify a string or buffer as the input types.

updated.

diversemix: I would not specify as bytes as this will result in the string not being encoded. I believe the best is to use a string which will be utf-8 encoded, see pep 3120 https://www.python.org/dev/peps/pep-3120/

PR #54, opened 2019-10-01
send put request to details service

GiancarloFusiello:
In reference to https://github.com/libero/publisher/issues/245

Adds a new task in `process_zip_dag.py` that compares categories found in the JATS xml to those in the details service. If a category is not found then it is added to the details service. Finally, the details service article resource is updated relating the relevant category ids.

### DIFF
--- a/dags/process_zip_dag.py
+++ b/dags/process_zip_dag.py
@@ -238,7 +238,7 @@ def send_article_info_to_details_service(**context) -> bytes:
     category_ids = []
     for category in categories_in_xml:
         response = requests.get(
-            '%s/catgories?name=%s' % (DETAILS_SERVICE_URL, quote_plus(category))
+            '%s/categories?name=%s' % (DETAILS_SERVICE_URL, quote_plus(category))
         )
         response.raise_for_status()
 
--- a/dags/process_zip_dag.py
+++ b/dags/process_zip_dag.py
@@ -10,6 +10,7 @@ from io import BytesIO
 from pathlib import Path
 from tempfile import TemporaryFile, NamedTemporaryFile
 from typing import List
+from urllib.parse import quote_plus
 from zipfile import ZipFile
 
 import requests
@@ -41,6 +42,7 @@ SERVICE_URL = configuration.conf.get('libero', 'service_url')
 TEMP_DIRECTORY = configuration.conf.get('libero', 'temp_directory_path') or None
 SEARCH_URL = configuration.conf.get('libero', 'search_url')
 WORKERS = configuration.conf.get('libero', 'thread_pool_workers') or None
+DETAILS_SERVICE_URL = configuration.conf.get('libero', 'details_service_url')
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


dags/process_zip_dag.py
```diff
@@ -228,6 +230,38 @@ def add_missing_uri_schemes(**context) -> bytes:
...
 
+def send_article_info_to_details_service(**context) -> bytes:
+    article_xml = get_article_from_previous_task(context)
+    article_id = jats.get_article_id(article_xml)
+
+    categories_in_xml = jats.get_categories(article_xml)
+    category_ids = []
+    for category in categories_in_xml:
+        response = requests.get(
+            '%s/catgories?name=%s' % (DETAILS_SERVICE_URL, quote_plus(category))
```
giorgiosironi: typo in `catgories`?

dags/process_zip_dag.py
```diff
@@ -228,6 +230,38 @@ def add_missing_uri_schemes(**context) -> bytes:
...
+        results = response.json().get('items')
+        if results:
+            category_ids.append(results[0]['id'])
+        else:
+            response = requests.post(
+                '%s/categories' % DETAILS_SERVICE_URL,
+                json={'name': category}
+            )
+            response.raise_for_status()
+            category_ids.append(Path(response.headers['Location']).stem)
```
giorgiosironi: should not assume a particular path format, following the `Location` header leads to a JSON body with a known schema; that is the place to get the ID from
(this makes the API chatty, but it's a trade-off for extensibility as paths and the service on the other side can be changed as long as they conform to the schema)

dags/process_zip_dag.py
GiancarloFusiello: updated.

dags/process_zip_dag.py
GiancarloFusiello: What would be the value of the `Location` header in this instance in your view?

giorgiosironi: Empirically an absolute path but not necessarily containing an hostname:
```
/categories/42
```
or
```
/categories/42/chemistry
```
in a possible alternate implementation.

GiancarloFusiello: Currently it is in the form of:
```
/categories/42
```
which `pathlib` deals with nicely. The `.stem` property returns all the text after the last `/`.

diversemix:
I'm intrigued to know what the "details service" is?

dags/process_zip_dag.py
```diff
@@ -396,7 +437,8 @@ def send_post_request_to_reindex_search_service():
 add_missing_jpeg_extensions.set_downstream(strip_related_article_tags)
 strip_related_article_tags.set_downstream(strip_object_id_tags)
 strip_object_id_tags.set_downstream(add_missing_uri_schemes_task)
-add_missing_uri_schemes_task.set_downstream(wrap_article)
+add_missing_uri_schemes_task.set_downstream(send_article_info)
+send_article_info.set_downstream(wrap_article)
```
giorgiosironi: not sure whether the article or its details should be sent first; the `content-store` and `scholarly-content-detail` are decoupled from each other so there is no strong dependency. Conceptually I would think the article should appear before being associated with a category, but it's not a strong requirement.

giorgiosironi:
> I'm intrigued to know what the "details service" is?

https://github.com/libero/scholarly-content-detail whose responsibility is to model the association of scholarly content (articles mainly) with research categories, article types, etc. for querying purposes such as producing listings.

giorgiosironi:
Conflicts, outdated due to new Hypermedia API approach, target service not deployed in real environments; closing.

PR #55, opened 2019-10-08
add documentation

GiancarloFusiello:
In reference to https://github.com/libero/publisher/issues/272

Adds documentation about the implementation of Apache Airflow for Libero Publisher.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,11 @@ from airflow import configuration
 SEARCH_URL = configuration.conf.get('libero', 'search_url')
 ```
 
+Apache Airflow uses the [boto library](https://boto3.amazonaws.com/v1/documentation/api/latest/index.html)
+to interface with AWS services using their AWSHook. Refer to the [boto library
+documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/quickstart.html#configuration) 
+regarding configuration relating to connecting to AWS services.
+
 ### Tests
 Tests are run using [pytest](https://pytest.org/en/latest/). Test files are 
 located in the `tests/` directory. Files needed for testing, such as archives and 
--- a/README.md
+++ b/README.md
@@ -3,17 +3,27 @@
 This project is an implementation of Libero JATS Ingester.
 
 Contents:
- - [Dependencies](#dependencies)
- - [Getting Started](#getting-started)
- - [Running Tests Locally](#running-tests-locally)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -34,10 +43,117 @@ AWS S3
...
+```
+
+### Tests
+Tests are run using [pytest](https://pytest.org/en/latest/). Test files are 
+located in the `tests/` directory. Files needed for testing, such as archives and 
+xml files, are placed in the `tests/assets/` directory. Before tests are run an
+sqlite database is used and initialized as some operations require database
+access.
+ 
+Some operators such as the `PythonOperator` expect a callable to be parsed to
```
giorgiosironi: can you add an example of another Operator (doesn't matter which one)?

GiancarloFusiello: Why?

GiancarloFusiello: This section is in reference to the context object during testing.

README.md
```diff
@@ -34,10 +43,117 @@ AWS S3
...
+[DAG composition](https://airflow.apache.org/concepts.html#bitshift-composition).
+
+![Simple Apache Airflow Architectural Diagram](https://miro.medium.com/max/1140/1*u6duhZD2J_i1zZ0Txq26Cg.png)
+
+### DAGs
+Airflow inspects the `dags/` folder for files that instantiate the DAG class.
+Airflow expects to find this directory in the `AIRFLOW_HOME` directory.
+
+### Configuration
+Configuration files can be found in the `config/` directory. Airflow expects to
```
giorgiosironi: can you document how [environment variables can be used](https://github.com/libero/sample-configuration/blob/master/.docker/jats-ingester/airflow.cfg#L6) in there?

GiancarloFusiello: The purpose of this documentation is not to detail all of Airflows features. Those can be found in the official documentation. This documentation provides details about this specific implementation.

giorgiosironi:
[`docker-compose` 1.23.2 CHANGELOG](https://github.com/docker/compose/blob/master/CHANGELOG.md#1232-2018-11-28):

> Reverted a 1.23.0 change that appended random strings to container names created by docker-compose up, causing addressability issues. Note: Containers created by docker-compose run will continue to use randomly generated names to avoid collisions during parallel runs.

README.md
giorgiosironi: We don't use other Operators I guess, so fine as it is.

PR #57, opened 2019-10-09
Organise libero package

GiancarloFusiello:
This PR provides a little house keeping.

- Move `aws.py` and `task_helpers.py` to the `libero` package
- Rename `task_helpers.py` to `context_facades.py`


tests/test_task_helpers.py
```diff
@@ -1,6 +1,6 @@
 import pytest
```
giorgiosironi: shouldn't this file be renamed to `test_context_facades.py` or `test_libero_context_facades.py`?

giorgiosironi:
(Working on https://github.com/libero/jats-ingester/pull/56 for the build problem)

tests/test_task_helpers.py
diversemix: surely `libero` is redundant in any filename?

GiancarloFusiello: Good catch. Renamed to `test_context_facades.py`.

PR #58, opened 2019-10-09
Grep name from docker containers list rather than hardcode it

GiancarloFusiello:


.travis/smoke-test.sh
```diff
@@ -8,6 +8,8 @@ function finish {
 
 trap finish EXIT
 
+name=$(docker ps | grep airflow_webserver | awk '{print $1}')
+
 docker-compose up -d
-.scripts/docker/wait-healthy.sh jatsingester_airflow_webserver_1 60
```
diversemix: If its important that you wait for the container to become healthy - why isn't this a part of the start command of the container?

giorgiosironi: `docker-compose up -d` is the start command of the container, but it never waits for container health due to it being asynchronous?

diversemix: Ah, I thought the `command` would be synchronous - that explains it... so if this is the case, what is not resilient to the `jatsingester_airflow_webserver_1` not being available? Maybe its this that needs to be made more tolerant?

giorgiosironi: Every container (or server process in general) has a bootstrap time between when it is started and when it opens a port listening for connections, serving them successfully; sometimes it's even ready as soon as the port is opened. In general however, this is what Readiness probes are for, waiting for a newly launched container to be ready before actually using it.

diversemix: Yeah this is an issue - the determination of readiness is the responsibility of the caller/client - not an external script.

giorgiosironi: Let's discuss separately (as Kubernetes will have an influence on all of this)

giorgiosironi:
https://github.com/libero/jats-ingester/pull/56 should fix all builds if `master` is merged into a branch, making this workaround unnecessary; but thanks for attempting to keep the build green which is always important to ensure changes can be made.

PR #59, opened 2019-10-18
replace strip_related_article_tags task with javascript equivalent

GiancarloFusiello:
In reference to https://github.com/libero/publisher/issues/270

- install nodejs version 10 (latest LTS) on Airflow instances
- create concept for running tasks written in javascript as Airflow tasks
  -  uses Airflow's BashOperator to call node with relevant scripts
  -  uses S3 to handle state between tasks by storing task return values in the newly created bucket 'jats-ingester-completed-tasks`
  -  the `function-caller.js` script deals with Airflow abstractions so that adopters only need to write task code rather than accompanying glue code.

As there are quite a lot of changes here, I will refactor some of the new code into smaller pieces in a future PR.


dags/js/aws/get-s3-client.js
```diff
@@ -0,0 +1,22 @@
+const AWS = require('aws-sdk');
+
+function getS3Client() {
+
+  let s3ConfigParams = {
+    apiVersion: '2006-03-01',
+    maxRetries: 3,
+    s3ForcePathStyle: true,
```
GiancarloFusiello: @giorgiosironi assumed this would be host style, so he will investigate.

giorgiosironi: I would expect that this is required by MinIO rather than S3, as AWS is deprecating path style:
https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/

giorgiosironi: Correction: would expect this is required for all fake implementations of S3.

giorgiosironi: @GiancarloFusiello to verify `s3ForcePathStyle: false` would work with a real S3; use local tests and personal credentials to do it

PR #60, opened 2019-10-18
add git lfs information to documentation

GiancarloFusiello:


README.md
```diff
@@ -22,6 +23,14 @@ Contents:
...
 
 * [Docker](https://www.docker.com/)
+* [Git LFS](https://git-lfs.github.com/)
+
+### Before getting started
+In order to use asset files (zip files, xml files, etc), for testing or to run
+the project locally, make sure you have [Git LFS](https://git-lfs.github.com/) 
+installed as the `tests/assets/` will not only contain a representation of files
+rather than the actual files. [Git LFS](https://git-lfs.github.com/) will take
+care of downloading/uploading large files.
```
giorgiosironi: I assume you don't need to run `git lfs something` on the first clone, as there should be a Git hook that runs it automatically :thinking:

GiancarloFusiello: I think if you already have it installed you don't need to run anything when cloning. If you clone without it being installed then you have to run `git lfs checkout` to download the files after installing git lfs.

PR #62, opened 2019-10-23
update completed tasks key format

GiancarloFusiello:


tests/js/function-caller.test.js
```diff
@@ -105,14 +94,20 @@ describe('Test functionCaller', () => {
...
     callable.mockReturnValue(undefined);
+    console.log = jest.fn();
 
-    expect.assertions(2);
     await functionCaller();
+
     expect(callable.mock.calls.length).toBe(1);
-    expect(AWS.S3.upload.mock.calls.length).toBe(0);
+
+    let logCallsLength = console.log.mock.calls.length - 1;
```
giorgiosironi: is it the length or is the the last call? `lastLogCallIndex`?

tests/js/function-caller.test.js
```diff
@@ -105,14 +94,20 @@ describe('Test functionCaller', () => {
 
 
   test('will not return AWS S3 Key if return value from callable not supplied', async () => {
+    // TODO: check S3 response object
+
+    let testKey = 'test/key';
 
-    AWS.S3.upload = jest.fn();
```
GiancarloFusiello: I removed this because it didn't work as originally intended. Left a TODO and used aws-sdk-mock instead.

tests/js/function-caller.test.js
GiancarloFusiello: good call. Updated.

PR #63, opened 2019-10-23
abstract bash operator with node to function

GiancarloFusiello:


dags/libero/operators.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def create_node_task(name: str,
+                     js_task_script_path: str,
+                     dag: DAG,
+                     xcom_pull: bool = False) -> BashOperator:
+
+    bash_command = 'nodejs {{ params.js_function_caller}} {{ params.js_task_script }}'
+    if xcom_pull:
+        bash_command += ' {{ ti.xcom_pull() }}'
```
giorgiosironi: pedantic and I know we are following the names from the original `BashOperator`, but this is more a `bash_command_template` than a `bash_command` as it itsn't executable without rendering it first

GiancarloFusiello: updated.

PR #64, opened 2019-10-23
update completed tasks key format in python task

GiancarloFusiello:


GiancarloFusiello:
> (all temporary code I hope)

Yes until other python tasks have been moved to JS.

PR #65, opened 2019-10-31
Convert python tasks to javascript

GiancarloFusiello:
In reference to https://github.com/libero/publisher/issues/281


dags/js/function-caller.js
```diff
@@ -87,5 +87,10 @@ module.exports = functionCaller;
 // if this script has been called directly then run the code in the if block
 // equivalent to python's __name__ == '__main__' comparison
 if (!module.parent) {
-  functionCaller();
+  try {
+    functionCaller();
```
giorgiosironi: `functionCaller` is `async` but we can't `await` here since we are at the top level of the module. The `catch` may not see errors then, will have to cross-check the approach

dags/js/tasks/convert-tiff-images-in-expanded-bucket-to-jpeg-images.js
```diff
@@ -0,0 +1,71 @@
+const fs = require('fs');
+const sharp = require('sharp');
+const uuidv4 = require('uuid/v4');
+const getS3Client = require('../aws/get-s3-client');
+
+async function convertTiffImagesInExpandedBucketToJpegImages() {
+
+  function deleteFile(fileName) {
+    fs.unlinkSync(fileName);
```
giorgiosironi: to be discussed: use of sync IO in production code

dags/js/tasks/extract-archived-files-to-bucket.js
```diff
@@ -0,0 +1,63 @@
...
+  s3Params = {Bucket: process.env.DESTINATION_BUCKET};
+
+  for (let entry of archive.getEntries()) {
+    if (entry.isDirectory) {
+      continue;
+    }
+
+    // TODO: stream data to s3 asynchronously
+    s3Params.Key = process.env.ARCHIVE_FILE_NAME.replace(new RegExp('\\.\\w+$'), '') + '/' + entry.entryName;
+    s3Params.Body = entry.getData();
```
giorgiosironi: is `getData` the synchronous part?

dags/js/function-caller.js
GiancarloFusiello: Can confirm that errors are definitely caught here. You can try this locally by throwing an error somewhere in the code.

dags/js/tasks/convert-tiff-images-in-expanded-bucket-to-jpeg-images.js
GiancarloFusiello: I can update this.

dags/js/tasks/extract-archived-files-to-bucket.js
GiancarloFusiello: yes, it returns a Buffer. I would prefer to add this to a file stream asynchronously.

dags/js/tasks/update-tiff-references-to-jpeg-in-article.js
```diff
@@ -0,0 +1,60 @@
...
+  // a few newlines (\n) can still be found
+  let xmlDocAsString = xmlDoc.toString(false);
+
+  // actually remove formatting
+  xmlDocAsString = xmlDocAsString.replace(new RegExp('\n', 'g'), '');
+
+  return Buffer.from(xmlDocAsString);
+}
+
+module.exports = updateTiffReferencesToJPEGInArticle;
```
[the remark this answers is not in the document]
GiancarloFusiello: Technically it would make no difference but when you have many open at once that look similar, it can be a helpful reminder. Either way, would refactor in another PR.

dags/js/tasks/convert-tiff-images-in-expanded-bucket-to-jpeg-images.js
GiancarloFusiello: updated to async version.

diversemix:
@GiancarloFusiello ... I suggest we go through this as part of the hand over tomorrow morning.

/cc @giorgiosironi @erezmus @Cooryd @will-byrne @nlisgo @thewilkybarkid

giorgiosironi:
Good starting point as it goes through all the various layers.

PR #70, opened 2019-11-07
Investigate and reduce race conditions

giorgiosironi:
For https://github.com/libero/publisher/issues/295

Transitions a mix of callbacks and promises to only promises.

### DIFF
--- a/dags/js/IO/file-utils.js
+++ b/dags/js/IO/file-utils.js
@@ -1,12 +1,11 @@
 const fs = require('fs');
+const util = require('util');
 
 
 module.exports.deleteFile = async (filePath) => {
   if (fs.existsSync(filePath)) {
-    await fs.unlink(filePath, (error) => {
-      if (error) throw error;
-      console.log('successfully deleted ' + filePath);
-    });
+    await util.promisify(fs.unlink)(filePath);
+    console.log('successfully deleted ' + filePath);
   } else {
     console.log(filePath, 'does not exist. Cancelling call to delete file.')
   }
--- a/dags/js/IO/file-utils.js
+++ b/dags/js/IO/file-utils.js
@@ -3,7 +3,7 @@ const util = require('util');
 
 
 module.exports.deleteFile = async (filePath) => {
-  if (fs.existsSync(filePath)) {
+  if (await util.promisify(fs.exists)(filePath)) {
     await util.promisify(fs.unlink)(filePath);
     console.log('successfully deleted ' + filePath);
   } else {
--- a/dags/js/aws/s3-utils.js
+++ b/dags/js/aws/s3-utils.js
@@ -4,11 +4,9 @@ const io = require('../IO/file-utils');
 const uuidv4 = require('uuid/v4');
 
 
-function createBucket(s3Params) {
+async function createBucket(s3Params) {
   let s3 = this.getS3Client();
-  s3.createBucket(s3Params, (error) => {
-    if (error) throw error;
-  });
+  await s3.createBucket(s3Params).promise();
 }
 
 async function getObject(s3Params) {
--- a/dags/js/aws/s3-utils.js
+++ b/dags/js/aws/s3-utils.js
@@ -78,9 +78,7 @@ async function upload(s3Params) {
   let s3 = this.getS3Client();
 
   console.log('Uploading:', s3Params.Bucket, s3Params.Key);
-  let response = await s3.upload(s3Params, (error) => {
-    if (error) throw error;
-  }).promise();
+  let response = await s3.upload(s3Params).promise();
   console.log(response.Location);
   return response;
 }
--- a/dags/js/aws/s3-utils.js
+++ b/dags/js/aws/s3-utils.js
@@ -12,11 +12,7 @@ async function createBucket(s3Params) {
 async function getObject(s3Params) {
   let s3 = this.getS3Client();
   console.log('Getting the following from S3: ', s3Params);
-  let response = await s3.getObject(s3Params, (error) => {
-    if (error) {
-      throw error;
-    }
-  }).promise();
+  let response = await s3.getObject(s3Params).promise();
   console.log('Data received from S3: ', response);
   return response;
 }
--- a/dags/js/aws/s3-utils.js
+++ b/dags/js/aws/s3-utils.js
@@ -30,13 +30,12 @@ async function getObjectToFile(s3Params) {
     let writable = fs.createWriteStream(tempDownloadFileName);
     writable.on('close', () => {
       console.log('Writable stream closed');
+      console.log(tempDownloadFileName, 'File size on writable close:', fs.statSync(tempDownloadFileName).size);
       resolve();
     });
     s3.getObject(s3Params).createReadStream()
       .on('end', () => {
-        // hypothesis: this is performed when the readable stream has been completely read, but the writable stream
-        // may still be written
-        console.log(tempDownloadFileName, 'File size on end:', fs.statSync(tempDownloadFileName).size);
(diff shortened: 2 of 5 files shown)


dags/js/IO/file-utils.js
```diff
@@ -1,12 +1,11 @@
 const fs = require('fs');
+const util = require('util');
 
 
 module.exports.deleteFile = async (filePath) => {
   if (fs.existsSync(filePath)) {
```
davidcmoulton: Better async?

dags/js/aws/s3-utils.js
```diff
@@ -4,21 +4,15 @@ const io = require('../IO/file-utils');
...
+async function createBucket(s3Params) {
   let s3 = this.getS3Client();
-  s3.createBucket(s3Params, (error) => {
-    if (error) throw error;
-  });
+  await s3.createBucket(s3Params).promise();
 }
 
 async function getObject(s3Params) {
   let s3 = this.getS3Client();
```
davidcmoulton: Better as `const` if not mutated (generally throughout).

docker-compose.tests.yml
```diff
@@ -65,6 +65,7 @@ services:
       - ./config/airflow.test.cfg:/airflow/airflow.cfg
     # TODO: create script for next line
     command: /bin/bash -c "/wait && sleep 3 && node node_modules/.bin/jest /airflow/tests/js/integration/"
+    #command: /bin/bash -c "/wait && sleep 3 && node node_modules/.bin/jest /airflow/tests/js/integration/extract-archived-files-to-bucket.test.js -t elife-00666"
```
nlisgo: Add a comment to explain the value.

tests/js/integration/extract-archived-files-to-bucket.test.js
```diff
@@ -31,9 +31,10 @@ describe('Test extractArchivedFilesToBucket', () => {
     process.env.ARCHIVE_FILE_NAME = 'elife-00666-vor-r1.zip';
     fu.deleteFile = jest.fn();
 
-    s3Utils.createBucket({Bucket: destinationBucket});
+    await s3Utils.createBucket({Bucket: destinationBucket});
 
 
+    // test seems to be declared green here, even if the rest is executed
```
nlisgo: Perhaps remove this now, as it should be resolved.

tests/js/integration/convert-tiff-images-in-expanded-bucket-to-jpeg-images.test.js
```diff
@@ -62,7 +62,7 @@ describe('Test convertTiffImagesInExpandedBucketToJpegImages', () => {
     process.env.DESTINATION_BUCKET = destinationBucket;
     process.env.ARCHIVE_FILE_NAME = 'biorxiv-685172.meca';
 
-    s3Utils.createBucket({Bucket: destinationBucket});
+    await s3Utils.createBucket({Bucket: destinationBucket});
 
     await extractArchivedFilesToBucket();
     fu.deleteFile = jest.fn();
```
davidcmoulton: Could extract this and the same on line 40 into a `beforeEach()`

tests/js/integration/extract-archived-files-to-bucket.test.js
```diff
@@ -52,7 +53,7 @@ describe('Test extractArchivedFilesToBucket', () => {
     process.env.ARCHIVE_FILE_NAME = 'biorxiv-685172.meca';
     fu.deleteFile = jest.fn();
```
davidcmoulton: Extract into `beforeEach()`?

dags/js/aws/s3-utils.js
```diff
@@ -4,21 +4,15 @@ const io = require('../IO/file-utils');
...
   console.log('Getting the following from S3: ', s3Params);
-  let response = await s3.getObject(s3Params, (error) => {
-    if (error) {
-      throw error;
-    }
-  }).promise();
+  let response = await s3.getObject(s3Params).promise();
```
davidcmoulton: Is it necessary to wait for the object before returning? Might be more efficient to return the promise and refactor the caller accordingly.

giorgiosironi: Could try fancy `finally` or `then()` solutions to deal with the `console.log`, but leaving this for now.

tests/js/integration/convert-tiff-images-in-expanded-bucket-to-jpeg-images.test.js
giorgiosironi: Tried, but doesn't work as `fu.deleteFile` needs to be stubbed from this point onwards rather than for the whole method.

tests/js/integration/extract-archived-files-to-bucket.test.js
giorgiosironi: Here we can.

docker-compose.tests.yml
giorgiosironi: :+1:

giorgiosironi:
Made all the changes that are applicable to this refactoring, build green.

dags/js/IO/file-utils.js
```diff
@@ -1,12 +1,11 @@
...
+const util = require('util');
 
 
 module.exports.deleteFile = async (filePath) => {
-  if (fs.existsSync(filePath)) {
-    await fs.unlink(filePath, (error) => {
-      if (error) throw error;
-      console.log('successfully deleted ' + filePath);
-    });
+  if (await util.promisify(fs.exists)(filePath)) {
```
diversemix: Using `existsSync` will make the code more readable.

dags/js/IO/file-utils.js
```diff
@@ -1,12 +1,11 @@
...
+    await util.promisify(fs.unlink)(filePath);
```
diversemix: Ditto here, by using `unlinkSync`

dags/js/IO/file-utils.js
diversemix: For reference no one should be using `fs.exists` ... ref: https://nodejs.org/api/fs.html#fs_fs_exists_path_callback 

/cc @libero/node-js-reviewers

giorgiosironi: Have just gone away from `existsSync` because it's sync. Will switch to `fs.access` then.

dags/js/IO/file-utils.js
giorgiosironi: Blocking IO should not be a standard for production code?

diversemix: But `deleteFile` is async so that should be ok then?

dags/js/IO/file-utils.js
diversemix: Why is this such a problem inside an `async` function?

diversemix: I created a `gist` to help illustrate that its fine... https://gist.github.com/diversemix/60cfcfb4cd378744aa03004cc4799c42

giorgiosironi: Forked the gist to demonstrate how the sync filesystem calls should block the whole process including all outstanding promises:
https://gist.github.com/giorgiosironi/034b62b7e4d11e9285007a9b0a1e9717

dags/js/IO/file-utils.js
giorgiosironi: See other comments chain

dags/js/IO/file-utils.js
diversemix: That's because the function isn't marked as async ... try changing line 24...
from

giorgiosironi: Live debugging session on the Gist, agreed to cut out `*sync()` calls in general from this project as problematic.

PR #71, opened 2019-11-14
Do not add JATS namespace to elements that already have a different one

giorgiosironi:
Elements start without a namespace (for some reason) so we add that JATS namespace to all of them. However, some elements may belong to a different namespace like the [Access License and Indicators one](https://jats.nlm.nih.gov/publishing/tag-library/1.2d1/attribute/xmlns-ali.html) so that doesn't apply to them.

See https://github.com/libero/publisher/issues/295#issuecomment-553848190 where an incorrect element is produced named `<ali:jats:free_to_read/>`.

### DIFF
--- a/dags/js/tasks/wrap-article-in-libero-xml.js
+++ b/dags/js/tasks/wrap-article-in-libero-xml.js
@@ -19,10 +19,16 @@ function wrapArticleInLiberoXML(xmlStringBuffer) {
   xmlDoc.root().attr('xml:base', base);
 
   function addJATSPrefix(element) {
-    element.name('jats:' + element.name());
-    element.childNodes().forEach((child) => {
-      addJATSPrefix(child);
-    })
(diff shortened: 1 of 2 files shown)


de-code:
Just a note: all of the JATS XML that I have encountered so far do not have an XML Namespace for the JATS elements (just the DTD), just for some elements like the one you mentioned. That includes the eLife XML in the GitHub repo, BioRxiv XML, PubMed (N)XML. Adding an XML namespace might break downstream processes that expect there not to be one.

giorgiosironi:
> Adding an XML namespace might break downstream processes that expect there not to be one.

Good to know. We don't export towards external systems in Libero, but if we did we would probably have to start from the original or strip the namespace again?

giorgiosironi:
Will get the feedback of trial by fire.

